PR #8, opened 2016-09-10
User logging unittest

vesnikos:
An initial unittest that simulates the loggin procedure


pierotofy:
Looks good, thanks @vesnikos!

PR #12, opened 2016-09-12
added fixtures for tests, updated tests to use fixtures

vesnikos:
Added fixtures to provide test data. The fixtures provide two test users and three projects that can be loaded to the project

The test data can also be used for booting up the test server for UI testing:

> manage.py testserver test_users test_project

you can then login normally with the username testuser and password test1234


pierotofy:
Looks good, thanks!

PR #17, opened 2016-09-27
Use https protocol

dakotabenjamin:
There is a higher likelihood of `git://` being blocked by a proxy or firewall than `https://`


pierotofy:
Looks good, thanks!

PR #29, opened 2016-10-07
Added default group logic with model wise permissions, tests, api project list with proper permissions check

### LINKED ISSUE #27: Check if superuser already exists in start.sh
This will avoid errors messages from popping up in docker-compose.

pierotofy:
Fixes #27


dakotabenjamin:
When running these tests, will you tell me if I need to migrate the db or any special commands? I want to make sure I am testing comprehensively, and being new to Django I am worried I may miss something.

pierotofy:
So the tests are automatically executed in the travis build process. If you are testing them on your server, all you need to do is making sure to start the app via ./start.sh. That script contains all the commands that are necessary to sync up the DB. You'll see that it executes:

```bash
python manage.py makemigrations
python manage.py migrate
```

Which generate the migrations and apply them. 

To run all tests, you run:

```
python manage.py test
```

This will execute all tests across all modules (currently `app` and `nodeodm`).

Does this answer your question? If not let me know.

dakotabenjamin:
OK great.

pierotofy:
Thanks @dakotabenjamin !

PR #32, opened 2016-10-13
Added react library, refactored tests, task creation and permissions, image uploads

pierotofy:
This is an incremental series of additions that span the addition of the reactjs library, better tests (removal of fixtures), permission handling for tasks and a proof of concept of image uploads.


pierotofy:
@vesnikos would you be available to code review this?

vesnikos:
Yep just give me a couple of days for a change to review it

pierotofy:
Good points; the main issue I was having with fixtures, was that signals are not properly triggered when loading fixtures data during testing.

For example, when a new user is added to the system, the user is automatically assigned to the "Default" group (which is created automatically during boot).

When using a fixture to load test users, unless I specify in the fixture that I want them to be part of the "Default" group, they are not going to be assigned to the "Default" group (signals are skipped).

I'm not against fixtures per se, I just could not find another way to test the signals code (see `signals.py`) and the "Default" group creation (and a few other initialization stuff, see `boot.py`). Perhaps there's a better strategy to approach this?

vesnikos:
Hmm can you tell me which test was falling on which commit at your PL?

pierotofy:
Sure, if you go back to https://github.com/OpenDroneMap/WebODM/commit/550816ee785f7916a3a9066eab7e7d03a0033aa3, see `signals.py`:

```python
@receiver(signals.post_save, sender=User, dispatch_uid="user_check_default_group")
def check_default_group(sender, instance, created, **kwargs):
    if created:
        try:
            default_group = Group.objects.get(name="Default")
            instance.groups.add(default_group)
            instance.save()
            logger.info("Added {} to default group".format(instance.username))
        except:
            pass # Group "Default" is not available, probably loading fixtures at this moment...
```

You'll notice that when users are loaded via fixtures, the check_default_group signal is not triggered.

pierotofy:
After more reading on stack overflow, I think it's really possible I might have been initializing signals the wrong way.

The end result is that the test suite should test that:
 - A default group should be created if there's none
 - New users should be automatically assigned to the default group upon creation
 - The default group has always certain permissions (view_project, change_project, etc.)
 - A project owner should always receive all object level permissions for that project (see `app/models.py` @ def project_post_save)

pierotofy:
Also note I've made some changes to the initialization code since then, the latest code is available here: https://github.com/pierotofy/WebODM/tree/processing

PR #39, opened 2016-10-26
Processing

pierotofy:
Scheduler, API client refactoring (no more swagger), task processing


pierotofy:
@mojodna let me know if you would have time to do a code check here?

mojodna:
@pierotofy sure thing.

mojodna:
FWIW, the codebase I've been working on (with similar goals but slightly different constraints) is here: https://github.com/mojodna/posm-opendronemap-api

pierotofy:
Yup, I've checked it out a few days ago; very nice work! Let me know if you need a pair of eyes for code reviews there. 👍

app/scheduler.py
```diff
@@ -0,0 +1,84 @@
...
+logger = logging.getLogger('app.logger')
+scheduler = None
+
+def job(func):
+    """
+    Adds background={True|False} param to any function
+    so that we can call update_nodes_info(background=True) from the outside
+    """
+    def wrapper(*args,**kwargs):
+        if (kwargs.get('background', False)):
```
mojodna: This seems backwards. If it's not backgrounded, I would expect it to run on the main thread.

app/scheduler.py
```diff
@@ -0,0 +1,84 @@
...
+        tasks_mutex.release()
+
+    def process(task):
+        task.process()
+        task.processing_lock = False
+        task.save()
+
+    if tasks.count() > 0:
+        pool = ThreadPool(tasks.count())
+        pool.map(process, tasks)
```
mojodna: `pool.apply` instead of `.map` since you don't care about the results?

app/scheduler.py
pierotofy: So currently if you call:

```python
myfunc(background=True) # runs in background (not on main thread)
myfunc(background=False) # on main thread
myfunc() # on main thread
```

The second parameter to .get is the default value if 'background' is not provided in the dict. Perhaps I should add a comment, it's kind of difficult to read.

nodeodm/api_client.py
```diff
@@ -1,35 +1,41 @@
...
-            return func(self, *args, **kwargs)
-        return check
-
     def __init__(self, host, port):
         self.host = host
         self.port = port
 
-    @check_client
+    def url(self, url):
+        return "http://{}:{}{}".format(self.host, self.port, url)
```
mojodna: Use `urlparse.urlunparse` instead to prevent the port number from showing up if it's the default (and to support https if necessary)?

mojodna:
lgtm, though I haven't been following along too closely, so my read is more about the changes themselves.

app/scheduler.py
pierotofy: `apply` however does not allow me to pass an iterable, so I would have to make a for loop?

```python
for task in tasks:
    pool.apply_async(process, task)
```

app/scheduler.py
```diff
@@ -0,0 +1,84 @@
...
+        pool.close()
+        pool.join()
+
+def setup():
+    global scheduler
```
pierotofy: Scope wise they are both declared at the module level. But that's a good point, I'm not too happy about using global. I'll rewrite the code to initialize `scheduler`:

```python
logger = logging.getLogger('app.logger')
scheduler = BackgroundScheduler()
```

instead of

```python
logger = logging.getLogger('app.logger')
scheduler = None
```

nodeodm/api_client.py
pierotofy: Good idea. I don't think the port number will be hidden though: 

```python
 urlunparse(('http', 'google.com:80', 'url', '', '', ''))
# "http://google.com:80/url"
```

The args for urlunparse are  scheme, netloc, url, params, query, fragment = data (no port though :( ).

pierotofy: I'll add code to check that separately (port number).

pierotofy:
Thanks a lot! Very valuable feedback, much appreciate.

PR #40, opened 2016-11-03
Task List

pierotofy:
The following changes add task list display functionality, including an output console, more API communication with node-OpenDroneMap and beginning of action functionality (ability to interact with a task).

@mojodna would you be willing to help me review some of these changes? You a great job with the last one.


app/api/tasks.py
```diff
@@ -68,13 +111,13 @@ def create(self, request, project_pk=None):
                     for file in filesList]
 
         task = models.Task.create_from_images(files, project)
-        if task != None:
+        if not task is None:
```
mojodna: While you're golfing, how 'bout `if task is not None:`?

app/api/tasks.py
```diff
@@ -42,10 +49,45 @@ def get_and_check_project(self, request, project_pk, perms = ('view_project', ))
...
+            task = self.queryset.get(pk=pk, project=project_pk)
+        except ObjectDoesNotExist:
+            raise exceptions.NotFound()
+
+        task.pending_action = task.PendingActions.CANCEL
+        task.last_error = None
+        task.save()
+
+        # Call the scheduler (speed things up)
+        scheduler.process_pending_tasks(background=True)
```
mojodna: This is totally not actionable, but I can't shake the feeling that you may be going down a road where you're reinventing celery.

app/models.py
```diff
@@ -124,56 +139,90 @@ def create_from_images(images, project):
...
+        ready to be processed execute some logic. This could be communication
+        with a processing node or executing a pending action.
+        """
+
+        if self.processing_node:
+            # Need to process some images (UUID not yet set)?
+            if not self.uuid:
+                logger.info("Processing... {}".format(self))
+
+                images = [image.path() for image in self.imageupload_set.all()]
```
mojodna: Where / how are images (and intermediate / completed artifacts) stored so that processing nodes have access to them?

app/static/app/js/components/EditTaskPanel.jsx
```diff
@@ -38,35 +38,51 @@ class EditTaskPanel extends React.Component {
       }
 
       this.nodesRequest = 
-        $.getJSON("/api/processingnodes/?online=True", json => {
+        $.getJSON("/api/processingnodes/?has_available_options=True", json => {
```
mojodna: Consider using `fetch` (and its corresponding polyfill) to eliminate one of the reasons you might be depending on jQuery.

app/static/app/js/Console.jsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import './css/Console.scss';
+import './vendor/google-code-prettify/prettify';
+import './vendor/google-code-prettify/prettify.css';
+import update from 'react-addons-update';
+import Utils from './classes/Utils';
+
+class Console extends React.Component {
```
mojodna: I was playing around with [xterm.js](https://github.com/sourcelair/xterm.js) yesterday. It doesn't eliminate the need for this component, but it may be a nice addition, especially since it does a nice job of display ANSI escape codes (which are present in the ODM log output).

Improved auto-scrolling behavior is being tracked in https://github.com/sourcelair/xterm.js/pull/336 and is scheduled to land in next Monday's release.

app/static/app/js/components/ErrorMessage.jsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    close(){
+        this.setState({error: ""});
+    }
+
+    render(){
+        if (this.state.error !== ""){
+            return (
+                <div className="alert alert-warning alert-dismissible">
```
mojodna: I was having good success with [React Bootstrap](https://react-bootstrap.github.io/) last week. It helped clear up the handling of modals particularly.

app/static/app/js/components/ProjectListItem.jsx
```diff
@@ -265,6 +266,8 @@ class ProjectListItem extends React.Component {
             <span>Updating task information... <i className="fa fa-refresh fa-spin fa-fw"></i></span>
           : ""}
 
+          {this.state.showTaskList ? <TaskList ref={this.setRef("taskList")} source={`/api/projects/${this.props.data.id}/tasks/?ordering=-id`}/> : ""}
```
mojodna: Interesting use of `ref`s. This is inverse of how I'm used to using components, so I've been thinking through the implications. It's more explicit than flowing `props` down to children, though it does lose the `propTypes` validation that would occur when passing functions _to_ children (potentially triggering actions).

I've been happy with Redux and associated patterns in the past, but for my most recent React app I didn't bother with the wiring and am somewhat regretting the externalization of state.

(Irrelevant aside done ;-)

app/api/tasks.py
pierotofy: Nice catch! Sounds less like Yoda that way... :)

app/static/app/js/components/TaskList.jsx
```diff
@@ -1,23 +1,68 @@
...
 
   render() {
+    let message = "";
+    if (this.state.loading){
+      message = (<span>Loading... <i className="fa fa-refresh fa-spin fa-fw"></i></span>);
+    }else if (this.state.error){
+      message = (<span>Could not get tasks: {this.state.error}. <a href="javascript:void(0);" onClick={this.refresh}>Try again</a></span>);
+    }else if (this.state.tasks.length === 0){
+      message = (<span>This project has no tasks. Create one by uploading some images!</span>);
+    }
```
mojodna: Proposed alternative:

```javascript
switch (true) {
case this.state.loading: {
  // ...
  break;
}
case this.state.error != null: {
  // ...
  break;
}
// ...
}
```

app/static/app/js/components/TaskListItem.jsx
```diff
@@ -0,0 +1,260 @@
...
+
+    let expanded = "";
+    if (this.state.expanded){
+      let actionButtons = [];
+      const addActionButton = (label, className, icon, onClick) => {
+        actionButtons.push({
+          className, icon, label, onClick
+        });
+      };
+      const genActionApiCall = (action) => {
```
mojodna: Move this out of `render()` (into `componentDidMount`?) to avoid creating lots of closures that do the same thing?

mojodna: Or into an instance method.

app/static/app/js/main.jsx
```diff
@@ -3,5 +3,17 @@ import './django/csrf';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import Dashboard from './Dashboard';
+import Console from './Console';
+import $ from 'jquery';
 
-ReactDOM.render(<Dashboard/>, document.getElementById('dashboard-app'));
+$("[data-dashboard]").each(function(){
```
mojodna: Ah, now I understand the jQuery. It's not a single React app, more multiple React fragments rendered together?

mojodna:
`genActionApiCall` is the only thing remotely resembling a yellow flag. Looks good to me.

app/api/tasks.py
pierotofy: In essence it is replicating some of the Celery features (it is a task processing queue). What is driving me to re-implement this mechanism is that so far, the task processing involves more than a single processing step (think input images in ---> assets out). Maybe I'm over-thinking it, but so far I found these cases where Celery does not seem like it would fare so well.

1. Input images in --> assets are not what I expected. I should be able to restart the process without re-sending the images, just by changing the processing options?

2. Console output: how to retrieve real-time console output from the opendronemap process?

3. How to protect processing nodes (authentication) without making it difficult to setup? Ideally a processing node should be (optionally) protected by a static token, or a userID/password combination that generates a temporary token. The docs here hint at using public/private key exchange for security: http://docs.celeryproject.org/en/latest/userguide/security.html (I'm not sure I'd want to go down that route)

app/models.py
pierotofy: `images` are paths to the local hard drive of the web server. They are saved by django in `app/media/project/<projectID>/task/<taskID>/filename.JPG`.

When a user uploads the images, they are stored there. Once all images have been uploaded, a processing node has been assigned and all options have been set, the images are sent to the processing node. The code you highlighted does this last step (sending the images to the processing node).

app/static/app/js/components/EditTaskPanel.jsx
pierotofy: I'll check it out! Thanks! I mainly use jQuery for the ajax calls.

app/static/app/js/Console.jsx
pierotofy: Nice find! I will keep an eye on it.

app/static/app/js/components/ErrorMessage.jsx
pierotofy: Yeah I had seen the project some time ago, but decided to pass on it because of this: ` The project is under active development, and APIs will change.` and this https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap (lots of stuff still a work in progress).

app/static/app/js/components/TaskList.jsx
pierotofy: Ah, neat! I've never seen the switch used that way. If I add a forth else if I will switch to that syntax (usually 3 if-else-if combinations are bearable to read).

app/static/app/js/main.jsx
pierotofy: Yes, correct! Since some parts of the app are powered by Django, certain parts are React apps, other parts use the Django forms. So far I think I'll have one app for the dashboard, one for the map display and one for the 3d model display. (Or maybe just a single app that does both map and 3d display).

Since I'm using jQuery this way, perhaps I'll leave `fetch` out of the picture for now.

pierotofy:
Great feedback, thanks @mojodna !

app/api/tasks.py
mojodna: Great, I was just thinking that.

PR #43, opened 2016-11-09
Python 3.5, PostGIS, assets download, map display

pierotofy:
This is a bulk update that moves the codebase to Python 3.5, adds PostGIS + GeoDjango db fields, asset downloads capabilities (including tile generation) and mockup of map display (which works, but is a work in progress).

@mojodna I cherry picked your Map.jsx code from posm admin 2 and made a few changes to make it work with the WebODM code base. Made me realize I've been missing out on using `defaultProps` for my components!

@vesnikos would you have a few minutes to review these changes? @mojodna you are also welcome as usual to give feedback!


pierotofy:
I'll merge these changes by Tuesday.

vesnikos:
Wow lot's of work has been put in this PL. 

Gdal is going to be installed by this commit which by turn will give access to a wide range of tools manipulating  geospatial data, ie the creation of footprints or thumbnails of rasters or can be easily be made and integrated in the project for additional value. Rasterio is a good example of what can be achieved.

pierotofy:
Yup, I think previews should be added at some point in the near future.

PR #50, opened 2016-11-18
Project actions

pierotofy:
This update brings in project actions (ability to create/delete projects), map display improvements, finer processing node permission settings, Swagger UI display.


pierotofy:
I'll merge this into master by Wednesday. Let me know if you have any feedback @mojodna !

mojodna:
I'll spin through in a bit.

app/api/common.py
```diff
@@ -0,0 +1,31 @@
+from django.core.exceptions import ObjectDoesNotExist
+from rest_framework import exceptions
+
+from app import models
+
+
+def get_and_check_project(request, project_pk, perms=('view_project',)):
+    '''
+    Retrieves a project and raises an exeption if the current user
```
mojodna: "exception"

app/api/common.py
```diff
@@ -0,0 +1,31 @@
...
+    has no access to it.
+    '''
+    try:
+        project = models.Project.objects.get(pk=project_pk, deleting=False)
+        for perm in perms:
+            if not request.user.has_perm(perm, project): raise ObjectDoesNotExist()
```
mojodna: Is there a "no access" in Django Core, or would that be layered on by some authorization framework? It seems like it might be appropriate to make that distinction here even if it's immediately caught and translated to a 404 (keeping 403s out of the mix makes sense to avoid leaking object existence).

app/api/common.py
```diff
@@ -0,0 +1,31 @@
...
+        raise exceptions.NotFound()
+    return project
+
+
+def get_tiles_json(name, tiles, bounds):
+    return {
+        'tilejson': '2.1.0',
+        'name': name,
+        'version': '1.0.0',
+        'scheme': 'tms',
```
mojodna: Have you tested this with something that understands TileJSON (e.g. [tessera](https://github.com/mojodna/tessera) with [tilejson](https://github.com/mapbox/node-tilejson): `tessera http://localhost/tiles.json`)?

If not, there's a good chance this should be `xyz`.

app/models.py
```diff
@@ -314,6 +331,18 @@ def process(self):
     def get_tile_path(self, z, x, y):
         return self.assets_path("orthophoto_tiles", z, x, "{}.png".format(y))
 
+    def get_tiles_json_url(self):
```
mojodna: Ditto.

app/models.py
```diff
@@ -314,6 +331,18 @@ def process(self):
...
+        return "/api/projects/{}/tasks/{}/tiles.json".format(self.project.id, self.id)
+
+    def get_tiles_json_data(self):
```
mojodna: Ditto

app/static/app/js/components/AssetDownloadButtons.jsx
```diff
@@ -1,48 +1,54 @@
...
                 <span className="caret"></span>
           </button>
           <ul className="dropdown-menu">
-            <li><a href="javascript:void(0);" onClick={this.downloadAsset("geotiff")}><i className="fa fa-map-o"></i> GeoTIFF</a></li>
-            <li><a href="javascript:void(0);" onClick={this.downloadAsset("las")}><i className="fa fa-cube"></i> LAS</a></li>
-            <li><a href="javascript:void(0);" onClick={this.downloadAsset("ply")}><i className="fa fa-cube"></i> PLY</a></li>
-            <li><a href="javascript:void(0);" onClick={this.downloadAsset("ply")}><i className="fa fa-cube"></i> CSV</a></li>
-            <li className="divider"></li>
-            <li><a href="javascript:void(0);" onClick={this.downloadAsset("all")}><i className="fa fa-file-archive-o"></i> All Assets</a></li>
+            {assetDownloads.map(asset => {
```
mojodna: `map`'s iterator includes a second argument you can use here, removing `i` from above:

```javascript
assetDownloads.map((asset, i) => {
```

app/static/app/js/components/Map.jsx
```diff
@@ -1,150 +1,179 @@
...
+                [info.bounds.slice(0, 2).reverse(), info.bounds.slice(2, 4).reverse()]
+              );
+            const layer = Leaflet.tileLayer(info.tiles[0], {
+                  bounds,
+                  minZoom: info.minzoom,
+                  maxZoom: info.maxzoom,
+                  tms: info.scheme === 'tms'
+                }).addTo(this.map);
+
+            // For some reason, getLatLng is not defined for tileLayer?
```
mojodna: I think it's attached to the map (Leaflet) instance.

mojodna:
No blockers, but a couple questions. Looks great!

app/static/app/js/components/AssetDownloadButtons.jsx
pierotofy: Ah, nice! It seemed awkward to initialize i outside.

app/static/app/js/components/Map.jsx
pierotofy: I added that because calling layer.openPopup() calls getLatLng on the layer so it knows where to place the popup.

app/static/app/js/components/Map.jsx
```diff
@@ -1,150 +1,179 @@
...
+            layer[Symbol.for("meta")] = meta;
+
+            this.imageryLayers.push(layer);
+
+            let mapBounds = this.mapBounds || Leaflet.latLngBounds();
+            mapBounds.extend(bounds);
+            this.mapBounds = mapBounds;
+
+            // Done loading all tiles?
+            if (++tilesLoaded === tiles.length){
```
pierotofy: Yes, good idea!

pierotofy:
Thanks for the helpful suggestions @mojodna !

app/api/common.py
mojodna: Ah, yes. I hadn't dug into how the tiles were created. The main difference is that TMS flips the Y coordinate, so the code is _slightly_ more complicated, which results in many implementations using xyz.

app/static/app/js/components/Map.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import '../css/Map.scss';
 import 'leaflet/dist/leaflet.css';
 import Leaflet from 'leaflet';
+import async from 'async/dist/async';
```
mojodna: Just `import async from 'async';` should work.

pierotofy: Ah, nice! Will land in master on the next PR, or perhaps you can change it. 👍

PR #52, opened 2016-11-26
Pagination

pierotofy:
This PR adds pagination functionality to the project list.

The components can be used to add pagination to other items by extending Paginated.jsx and using the Paginator component for rendering the buttons.

@mojodna would you be able to take a look sometimes?


app/static/app/css/main.scss
```diff
@@ -94,6 +94,40 @@ ul#side-menu.nav{
     z-index: 999999;
 }
 
+.pagination{
```
mojodna: _pushes for consistency of spacing when specifying rules_

app/static/app/js/components/Paginated.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import update from 'react-addons-update';
+
+/*abstract*/ class Paginated extends React.Component{
+  constructor(){
```
mojodna: Unnecessary.

app/static/app/js/components/Paginated.jsx
```diff
@@ -0,0 +1,84 @@
...
+    super();
+  }
+
+  updatePagination(itemsPerPage, totalItems){
+    let currentPage = 1;
+    const totalPages = this.totalPages(itemsPerPage, totalItems);
+
+    if (this.state.pagination && this.state.pagination.currentPage !== undefined){
```
mojodna: I typically use `!= null` (not `!==`) because it covers both the `null` and `undefined` cases.

app/static/app/js/components/Paginated.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+    this.setState({pagination: {
+        switchingPages: false,  
+        itemsPerPage: itemsPerPage,
+        totalItems: totalItems,
+        currentPage: currentPage
+      }
+    });
+
+    this.handlePageChange = this.handlePageChange.bind(this);
```
mojodna: Put this in the constructor instead?

app/static/app/js/components/Paginated.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import update from 'react-addons-update';
```
mojodna: From https://facebook.github.io/react/docs/update.html:

```
Note: update is a legacy add-on. Use kolodny/immutability-helper instead.
```

(I've never used either of these, which is why I went to go look it up.)

mojodna:
This is nice, I quite like the way you've implemented the pagination functionality.

app/static/app/js/components/Paginated.jsx
pierotofy: I tend to use null to indicate a variable that has been initialized to "nothing", while undefined for a variable that has not even been initialized. In this case my reason was to check whether pagination.currentPage has been initialized. Both approaches are fine and loose equality here would not not harm.

app/static/app/js/components/Paginated.jsx
pierotofy: Nice catch!

app/static/app/js/components/Paginated.jsx
mojodna: loljs, _sigh_.

PR #60, opened 2016-12-01
Install enhancements

pierotofy:
This PR aims to simplify the process of getting started with WebODM. A processing node is automatically setup for users so that they can start trying out the functionality of the software with minimal configuration. A python version check is also included in the start.sh script.

@dakotabenjamin would you be able to check these changes before I merge them in?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Install the following applications (if they are not installed already):
  - [Python](https://www.python.org/downloads/)
  - [Git](https://git-scm.com/downloads)
 
-* From the Docker Quickstart Terminal (Windows) or from the command line (OSX/Linux) type:
+* From the Docker Quickstart Terminal (Windows) or from the command line (OSX / Linux) type:
 ```
 git clone https://github.com/OpenDroneMap/WebODM
 cd WebODM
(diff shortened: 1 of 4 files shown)


dakotabenjamin:
yeah I'm on it

dakotabenjamin:
Running into this issue:
image

pierotofy:
I think you're getting that error because you have run WebODM without Docker before? It looks like the nodeodm/external/node-OpenDroneMap already contains some files.  Make sure that your linux host's nodeodm/external/node-OpenDroneMap directory is empty (or delete that folder altogether).

dakotabenjamin:
It's a fresh machine + in a virtual environment (because for some reason i was getting version errors with docker-compose)

dakotabenjamin:
I just ran the wrong docker-compose command. needed the -f tags.

pierotofy:
Ahh, that makes sense; just to confirm, you probably ran `docker-compose -f docker-compose.nodeodm.yml up`  instead of `docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up`?

dakotabenjamin:
No, I didn't look at the new readme (was gonna do the code review while its installing) and just rand `docker-compose up`

dakotabenjamin:
Works flawlessly, really impressed. 

The interface is really nice too, I hadn't gotten a chance to update until now.

pierotofy:
Thanks! It still needs a lot of work, but now it's more than just a mock, it actually does some processing, which should capture the interest of very early adopters beside just developers.

Issue #63, opened 2016-12-06
Create Docker Startup Script


pierotofy:
Some people are experiencing issues using Docker when rebuilding images.

A script should be created to handle the following tasks:

- Start WebODM
- Stop WebODM
- Rebuild From Scratch Images, ignoring cache and removing the proper volumes.

Issue #48, opened 2016-11-17
Generate better API docs


pierotofy:
Probably using https://github.com/lord/slate is a good option. Swagger is too picky and limited.

pierotofy:
Started working on this: https://pierotofy.github.io/WebODM/#introduction

pierotofy:
This is done

Issue #65, opened 2016-12-12
Error messages on install


leszekmp:
Installed the Docker image on Windows 10 64-bit. In the terminal window are repeated error messages:

[##mwebapp exit with code 127  (where ## is an integer in the range 31 through 36)
No such file or directory[0m /usr/ bin/env: bash
No such file or directory/usr/ bin/env: bash
No such file or directory/usr/ bin/env: bash
(repeated a total of 10 times)

Looking at the webapp container log in Kitematic, the "No such file or directory/usr/ bin/env: bash" also shows up 11 times.

Took a while, but figured out how to open up the WebApp. Am processing a simple task, and it seems to be working. So are those error messages of concern?

pierotofy:
Perhaps :)

I'm guessing this is the `wait-for-it.sh` script. Would you be able to change this line from:

```
#!/usr/bin/env bash
#   Use this script to test if a given TCP host/port are available
```
To:

```
#!/bin/bash
#   Use this script to test if a given TCP host/port are available
```

And report whether the problem still happens?

leszekmp:
Thanks for the fast response. Not that up on Linux, but I'll try to figure out how to do that.

pierotofy:
You can do that from Windows, simply open the `wait-for-it.sh` script with Notepad (it's in the WebODM root directory), do the changes and rebuild the docker image using Kinematic.

leszekmp:
Took a while, but I figured out how to do that. Nope - I now get the following error message in the logs for the web app:

/bin/bash: /webodm/wait-for-it.sh: /bin/bash^M: bad interpreter: No such file or directory

in addition to the previous error message.

I am changing the script back to its previous form.

pierotofy:
Thank you for trying!

Would you be able to copy/paste the whole output of the terminal window so that I can try to identify the place where this error is happening?

leszekmp:
This is not exactly intuitive for a Windows guy :). Here's the output from the Docker terminal window, after I deleted the original Containers.

$ docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up
Creating db
Creating node-odm-1
Creating webapp
Attaching to db, node-odm-1, webapp
[36mdb            |[0m The files belonging to this database system will be owned by user "postgres".
[36mdb            |[0m This user must also own the server process.
[36mdb            |[0m
[33mnode-odm-1    |[0m info: No tasks dump found
[...]

PR #66, opened 2016-12-13
Task stuck fix

### LINKED ISSUE #55: Task stuck on running
Processing the odm_data_bellus dataset with --resize-to 1800 option, the task remains stuck on running. node-OpenDroneMap successfully processed the images.

pierotofy:
This is an update to fix #55, #63 and start #48. It also contains minor CSS fixes and better error handling during processing.

@mojodna any feedback is greatly appreciated! As you can see I added support for offdb rasters in Django via a new "OffdbRasterField", which replaces the classical "RasterField".

I plan to merge in these changes by Thursday.


mojodna:
Thanks for giving me a window ;-) Trying to keep on top of things.

app/postgis.py
```diff
@@ -0,0 +1,218 @@
+import binascii
```
mojodna: What about... (see, I'm still stuck) http://postgis.net/docs/RT_ST_AsBinary.html in which case it really is just raw bytes.

> By default, WKB output contains the external file path for out-db bands. If the client does not have access to the raster file underlying an out-db band, set `outasin` to TRUE.

I get that it's convenient for a client to get the filename and read the raster itself, but you're not doing that here, are you?

slate/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+If this is a question or feature request, make sure to:
```
mojodna: Can Slate be added as a submodule? Does that make sense?

start.sh
```diff
@@ -9,22 +9,35 @@ echo "|__/|__/\___/_.___/\____/_____/_/  /_/   "
...
+python -c "import sys;ret = 1 if sys.version_info <= (3, 0) else 0;print('Checking python version... ' + ('3.x, good!' if ret == 0 else '2.x'));sys.exit(ret);"
+if [ $? -eq 1 ]; then
+	almost_there
     echo -e "\e[33mYour system is currently using Python 2.x. You need to install or configure your system to use Python 3.x. Check out http://docs.python-guide.org/en/latest/dev/virtualenvs/ for information on how to setup Python 3.x alongside your Python 2.x install.\e[39m"
     echo
     exit
 fi
 
+# Check GDAL version
+python -c "import sys;import re;import subprocess;version = subprocess.Popen([\"gdalinfo\", \"--version\"], stdout=subprocess.PIPE).communicate()[0].decode().rstrip();ret = 0 if re.compile('^GDAL [2-9]\.[1-9]+').match(version) else 1; print('Checking GDAL version... ' + ('{}, excellent!'.format(version) if ret == 0 else version));sys.exit(ret);"
```
mojodna: Uh... Why the Python?

```bash
gdalinfo --version | grep -qE '^GDAL [2-9]\.[1-9]+'
```

start.sh
```diff
@@ -9,22 +9,35 @@ echo "|__/|__/\___/_.___/\____/_____/_/  /_/   "
...
+if [ $? -eq 1 ]; then
```
mojodna: These should be `-ne 0`.

wait-for-it.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
mojodna: Better to use the old one in case `bash` is installed elsewhere (or a newer version is in one's `PATH`.

webodm.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+platform="Linux" # Assumed
+uname=`uname`
```
mojodna: I've found that `$(...)` is less error-prone than backticks (and possible to write in Markdown!).

webodm.sh
```diff
@@ -0,0 +1,81 @@
...
+if [[ $uname == "Darwin" ]]; then
```
mojodna: Perfect opportunity for a `case` statement!

webodm.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
```
mojodna: `set -eo pipefail` (`-o pipefail` is a bash-ism) is nice; makes things fail early and often (w/o `pipefail` errors sometime get masked).

webodm.sh
```diff
@@ -0,0 +1,81 @@
...
+	check_msg="\e[92m\e[1m OK\e[0m\e[39m"
+	hash $1 2>/dev/null || not_found=true 
+	if [[ $not_found ]]; then
+		check_msg="\e[91m can't find $1! Check that the program is installed before launching WebODM. $2\e[39m"
+	fi
+
+	echo -e "Checking for $1... $check_msg"
+	
+	if [[ $not_found ]]; then
+		exit 1
```
mojodna: Using `set -e` (above) means that this can `return 1` instead and be useful as a test, rather than something that always terminates.

app/migrations/9999_postgis.py
```diff
@@ -0,0 +1,9 @@
+from django.db import migrations
+
+class Migration(migrations.Migration):
+    dependencies = [("app", "0001_initial")]
+
+    operations = [
+        migrations.RunSQL("SET postgis.enable_outdb_rasters TO True;"),
+        migrations.RunSQL("SET postgis.gdal_enabled_drivers TO 'GTiff';")
```
pierotofy: Good catch, it actually is not persistent, I replaced it with a ALTER SYSTEM SET ... Thanks!!

README.md
```diff
@@ -22,21 +22,33 @@ If you know Python, web technologies (JS, HTML, CSS, etc.) or both, make a fork,
...
 pip install docker-compose
-docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up
+./webodm.sh start
 ```
 
-* If you're on Windows / OSX, find the IP of your Docker machine by running this command from your Docker Quickstart Terminal:
+* If you're on Windows find the IP of your Docker machine by running this command from your Docker Quickstart Terminal:
 
 ```
 docker-machine ip
```
pierotofy: So Linux / Mac uses its own IP address, Windows needs to figure it out via `docker-machine ip`. Was this more a suggestion to change "Linux / Mac" with "{Linux, Mac}"?

app/postgis.py
pierotofy: ST_AsBinary works, and I think the combination of ST_AsBinary and ST_AddBand,ST_CreateEmptyRaster would have allowed me to store offdb rasters. But it would have been a "hackhish" way to do it, plus I'm not sure it would have played 100% nice with Django's ORM. 

Here the from_pgraster and to_pgraster methods go from HEXWKB --> raster info (including the path of the offdb raster) and raster info --> HEXWKB. Once I have the path of the offdb raster I can initialize the GDALRaster object, which is the standard object in GeoDjango to reference raster fields.

I think I will ping the Django maintainers and see if they are interested in incorporating some of these changes into GeoDjango core.

slate/.github/ISSUE_TEMPLATE.md
pierotofy: It would be challenging to separate the configuration files and document definition files from the application's core files. Slate's instructions specifically say to make a full copy of the source and put it in your own repository.

start.sh
pierotofy: Ah, mostly because I'm more familiar with Python and less with Bash. :) I also call python -c a few lines above for another check. Perhaps somebody could rewrite it in Bash.

wait-for-it.sh
pierotofy: I changed this because somebody reported #65, and I know Windows does not include /usr/bin/env, so I'm trying to see if that is going to fix it.

webodm.sh
pierotofy: Nice! I did not know that. Good addition.

webodm.sh
pierotofy: Yes good idea!

webodm.sh
pierotofy: Nice! I changed it to return 1 instead.

slate/.github/ISSUE_TEMPLATE.md
mojodna: Gotcha. Following instructions is good ;-)

webodm.sh
```diff
@@ -0,0 +1,83 @@
...
+
+rebuild(){
+	run "docker-compose down"
+	run "rm -fr node_modules/"
+	run "rm -fr nodeodm/external/node-OpenDroneMap"
+	run "docker-compose build --no-cache"
+	echo -e "\e[1mDone!\e[0m You can now start WebODM by running ./$0 start"
+}
+
+if [[ $1 = "start" ]]; then
```
mojodna: Ooh, another good spot for a `case` statement. (Yes, too little, too late.) I'd submit a PR on my own, but I'm swamped at the moment.

webodm.sh
```diff
@@ -0,0 +1,83 @@
...
+check_command(){
+	check_msg="\e[92m\e[1m OK\e[0m\e[39m"
+	hash $1 2>/dev/null || not_found=true 
+	if [[ $not_found ]]; then
+		check_msg="\e[91m can't find $1! Check that the program is installed before launching WebODM. $2\e[39m"
+	fi
+
+	echo -e "Checking for $1... $check_msg"
+	
+	return 1
```
mojodna: `return false` also produces `1` and is clearer.

webodm.sh
pierotofy: Yes please, if you see anything that can be improved, make a PR.

start.sh
```diff
@@ -9,22 +9,35 @@ echo "|__/|__/\___/_.___/\____/_____/_/  /_/   "
...
-if [ $? -eq 1 ]; then
+almost_there(){
     echo 
     echo "===================="
     echo "You're almost there!"
     echo "===================="
+}
+
+# Check python version
+python -c "import sys;ret = 1 if sys.version_info <= (3, 0) else 0;print('Checking python version... ' + ('3.x, good!' if ret == 0 else '2.x'));sys.exit(ret);"
```
mojodna:
```bash
python --version 2>&1 | grep -qE '^Python 3'
```

PR #78, opened 2017-01-21
3D point cloud, textured model display, miscellaneous fixes

pierotofy:
This is a cumulative update that brings 3D point cloud and textured model display, as well as miscellaneous fixes and additions to the UI to allow users to navigate to the new areas of the UI.

@mojodna would you have some time to look over these changes? I plan to merge these by Tuesday.


pierotofy:
@dakotabenjamin feel free to look at it also. It's a large update, but most of the additions are from the inclusion of three.js / potree.

hobu:
Just a note to add that both [plasio.js](https://github.com/hobu/plasio.js), [greyhound](https://github.com/hobu/greyhound), and [entwine](https://entwine.io) may be interesting to use here as well. The soon-to-be-released potree has support for Greyhound. 

The advantage of Entwine is parallel builds and it supports any data schema -- not just LAS/LAZ. Greyhound, the web services piece (node.js) that utilizes Entwine-organized data to answer client requests, also supports user-defined data schemas.

app/static/app/js/ModelView.jsx
```diff
@@ -0,0 +1,1067 @@
...
+        
+        renderer.clearDepth();
+        measuringTool.render();
+        transformationTool.render();
+      };
+    };
+    var potreeRenderer = new PotreeRenderer();
+
+    // high quality rendering using splats
+    var highQualityRenderer = null;
```
mojodna: Move this above `loop()` so this doesn't feel like it's randomly hanging out?

app/static/app/js/ModelView.jsx
```diff
@@ -0,0 +1,1067 @@
...
+
+    function loop() {
+      requestAnimationFrame(loop);
+      update();
+      
+      if(quality === "Splats"){
+        if(!highQualityRenderer){
+          highQualityRenderer = new HighQualityRenderer();
+        }
+        highQualityRenderer.render(renderer);
```
mojodna: Does this always get initialized?

pierotofy: It gets initialized only if you select the "splats" display for points. I moved the reference to the `highQualityRenderer` closer to `loop` so hopefully it's more clear.

app/static/app/js/ModelView.jsx
pierotofy: Good idea!

dakotabenjamin:
After pulling the latest node-odm, I was able to view the 3D point cloud and textured model. It works really well! There is no option to only download the textured model, which doesn't seem intuitive since it is displayed.

pierotofy:
Good point @dakotabenjamin, I opened https://github.com/OpenDroneMap/WebODM/issues/80 for that.

desireB:
I was able to run Potree and display the 3D point cloud : 

1.Index with Entwine
2.Stream with Greyhound
3.View with Potree : https://github.com/connormanning/potree.entwine.io with Apache from Bitnami
` docker run -it -v $HOME:$HOME connormanning/entwine build -i /PATH_TO_LAS/odm_georeferenced_model.ply.las -o ~/greyhound/bendogo
`
`docker run -it -p 8080:8080 -v ~/greyhound:/greyhound connormanning/greyhound
`



Go to http://MACHINE-IP:8081/potree_entwine_io/data/custom.html?s=localhost:8080&r=bendogo

**Being able to download the textured model would be awesome indeed. It's really precise.**

bendogo

bendogo_cloud_point

pierotofy:
Cool! @desireB you can download the textured model by going to "Download Assets" --> "All".

ctrl6DB:
Thank you ! It was there indeed. Opened it with Meshlab, very cool.
Is there an option to generate more precise 3D cloud point ? What parameters would you recommend in Options -> Avanced settings for a high performance setup ?
Thanks.

desireB:
Thank you ! It was there indeed. Opened it with Meshlab, very cool.
Is there an option to generate more precise 3D cloud point ? What parameters would you recommend in Options -> Avanced settings for a high performance setup ?
Thanks.

pierotofy:
Try to enable the use-opensfm-pointcloud.

PR #95, opened 2017-02-08
Authtoken

pierotofy:
This PR adds support for access tokens. Clients can now authenticate via user/pass and receive a token to pass to subsequent API calls.

### DIFF
--- a/webodm/settings.py
+++ b/webodm/settings.py
@@ -11,6 +11,8 @@ https://docs.djangoproject.com/en/1.10/ref/settings/
 """
 
 import os, sys
+
+import datetime
 from django.contrib.messages import constants as messages
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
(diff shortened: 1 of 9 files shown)


webodm/settings.py
```diff
@@ -220,9 +222,18 @@
...
+  'DEFAULT_AUTHENTICATION_CLASSES': (
+    'rest_framework.authentication.SessionAuthentication',
+    'rest_framework.authentication.BasicAuthentication',
+    'rest_framework_jwt.authentication.JSONWebTokenAuthentication',
+  ),
   'PAGE_SIZE': 10,
 }
 
+JWT_AUTH = {
+    'JWT_EXPIRATION_DELTA': datetime.timedelta(hours=6),
```
dakotabenjamin: Is this arbitrary or recommended?

pierotofy: Depends on the paranoia level :)

Issue #101, opened 2017-02-11
refineBatchPFGS failed


AnthonyCoughlin:
WebODM stuck in a loop.  Watching Docker the command line is still running but has been repeating the same 'warnings' and WebODM is throwing an error message 'refineBatchPFGS failed'.  See attached.

![dog_refinepatch_failed](https://cloud.githubusercontent.com/assets/25381671/22858251/00057002-f07d-11e6-8197-a2d3882a321c.JPG)

![webodm_bfgs_failed](https://cloud.githubusercontent.com/assets/25381671/22858252/06ac3a6c-f07d-11e6-8cee-e736c218ef02.JPG)

![webodm_bfgs_failed2](https://cloud.githubusercontent.com/assets/25381671/22858268/73b9e744-f07d-11e6-8d59-1ed1eeadfa0d.JPG)

Let me know if you need more info!

AnthonyCoughlin:
Still going.

![webodm_bfgs_failed3](https://cloud.githubusercontent.com/assets/25381671/22858615/5be0e31a-f088-11e6-86f4-fb6eb2875bcd.JPG)

pierotofy:
It is possible that the ReadTimeout error stopped execution. I'll have to investigate to tell for sure. refinePatchesBFGS often times fails during processing, but should not interrupt the process.

Restarting the task might fix it temporarily.

AnthonyCoughlin:
@pierotofy  Thanks again for the assistance.  When this happened it corrupted the entire Docker db.  I've received db errors when trying to restart the Docker Machine.  I ended up wiping the WebODM folder again and uninstalling Docker Toolbox.  Then reinstalling Docker Toolbox and recloned WebODM.  I had to run `docker-compose build` to repair the db (not sure why I thought the uninstall would have removed all files) then ran `./webodm.sh update`.  Now WebODM starts with no errors.  However, I attempted to stitch the same dataset today (FYI - I uploaded the same data to MapsMadeEasy.com last night and it worked just fine - I'm pro ODM tho man!) and it failed.  I received this error (see below).  I am forcing focal-length and CCD width, I may be putting in the wrong values - would that cause an error?

![webodm_error_2-12-16](https://cloud.githubusercontent.com/assets/25381671/22866526/787f2792-f13d-11e6-9863-279547d87655.JPG)

AnthonyCoughlin:
[...]

PR #103, opened 2017-02-15
Timeout Handling, auto processing node selection

pierotofy:
Strengthened timeout error handling, added auto processing node selection mechanism on server side, automatic processing node reassignment, unit testing.

@mojodna would you be able to take a look at these changes if you have some time?

This PR also deals with the error messages shown in #101, but does not fix the root cause of that issue.


pierotofy:
Merged in, I will add modifications or suggestions as they come.

PR #111, opened 2017-03-01
Docs

pierotofy:
This is a cumulative update that mainly takes care of writing an initial complete documentation for WebODM's API.

Other changes include minor bug fixes, addition of a few API calls, additional unit testing.

@mojodna @dakotabenjamin @smathermather I plan to merge these changes in by Friday. If there's any feedback, please let me know.


smathermather:
Looking good overall.

I ran into issues with binding to the default address, and had to edit webodm/settings.py to add my IP to allowed hosts but it was not a clean install, so I don't know if this is an upgrade issue folks will have, or if I did something funky.

Also, I noticed ```use-opensfm-pointcloud``` still in the settings, but I expect this is that it is not pulling from the latest OpenDroneMap release, so its not yet default.

pierotofy:
@smathermather  were you using the latest master? `ALLOWED_HOSTS` has been set to [*] for a while now (allow all).

Yes the current master still pulls an older version of OpenDroneMap. This PR will use the newer version.

smathermather:
I thought I was pulling directly from the branch on this pull request. Let me check once my process is done... .

pierotofy:
Check that your processing node docker image is pulling from opendronemap/node-opendronemap and not from pierotofy/nodeodm. Alternatively, you can manually start a processing node using opendronemap/node-opendronemap and add it to WebODM. Newer options will be pulled in then.

pierotofy:
@smathermather actually, the opendronemap/node-opendronemap was last updated 5 days ago, so you probably won't see the new changes just yet. It will be available as soon as we rebuild that image.

app/api/processingnodes.py
```diff
@@ -35,4 +39,43 @@ class ProcessingNodeViewSet(viewsets.ModelViewSet):
...
+                    for option in node.available_options:
+                        if common_option['name'] == option['name']:
+                            found = True
+                            break
+
+                    # Mark for deletion
+                    if not found:
+                        common_option['_delete'] = True
+
+        common_options = [co for co in common_options if not '_delete' in co]
```
pierotofy: Update would allow me to create a union, but I'm looking to do an intersection. I don't think Python has anything built-in for that (especially since my list needs to be intersected based on a particular key within the list of objects).

I'm sure there is be a more compact way to write this however.

slate/source/includes/_quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+        if chunk:
+            f.write(chunk)
+print("Saved ./orthophoto.tif")
+```
+
+Our orthophoto is ready to be downloaded. A variety of other assets, including a dense 3D point cloud and a textured model [are also available](#download-assets).
+
+Congratulations! You just processed some images.
+
+![Success](https://i.imgflip.com/2/ipzhf.jpg)
```
pierotofy: Haha, I was hoping somebody would appreciate that.

dakotabenjamin: Does `dict.keys() & dict.keys()` do an intersection?

pierotofy:
@smathermather the new images have been built: https://hub.docker.com/r/opendronemap/node-opendronemap/

app/api/processingnodes.py
dakotabenjamin: Doesn't `dict.keys() & dict.keys()` do an intersection?

dakotabenjamin: How big are the two dicts? I'm just worried about the nested loops if they get too big

pierotofy: I thought about using `&`, but they aren't simple dicts, common_options and available_options are a list of dict objects.

[{'name': value}, {'name': value}, ...]

pierotofy: Size is under control, since its the number of command line options times the number of processing nodes.

So O(command line options ^ 2 * processing nodes).

But all numbers are pretty small.

dakotabenjamin: It probably won't be an issue then. What's the max number of nodes you've tested on? I wonder what kind of scaling problems we might run into overall with 100s of them!

pierotofy: Only about 5. I'm sure this information could be cached if performance became an issue.

smathermather:
So @pierotofy -- dumb question: if I build (after pointing to opendronemap/node-opendronemap), how do I use that new image in the webodm.sh script?

```
docker build -t mywebodm .
```

pierotofy:
@smathermather if you are using docker for every component (webodm, database and node-odm), you might want to just use:

`./webodm.sh rebuild`

Then go for a coffee while it rebuilds everything for you.

If you have built your own `mywebodm` image, you'll need to start and configure your database, start your node-odm image, then you can start the mywebodm image.

In general you don't want to manually issue a `docker build` command since docker-compose orchestrates the launch of WebODM and its dependencies. See `docker-compose.yml` and `docker-compose.nodeodm.yml`. Hope this makes sense.

smathermather:
This makes perfect sense. This is a new level of docker for me, so enjoying digging in.

Issue #115, opened 2017-03-08
webodm update deletes all projects in DB


hblanken:
Objective is to keep DB data persistent between updates. Currently the update command will nuke the db. I think this is an issue. Docker containers are not really designed to keep persistent data (although some people do it anyways). The best thing at the moment to persist data between updates would be do run WebODM / PostGIS on a local machine, and use docker to just run node-opendronemap. But there might be a way around it using docker volumes...or another solution to be investigated.

pierotofy:
After doing some initial research, it should be possible to maintain volume persistence by defining a named volume for the DB instance. 

@hblanken since you experienced this problem, are there any files left from your previous projects in ~/webodm/app/media/project ?

hblanken:
I might be looking in the wrong place - but my media folder is empty when I go through terminal.

hblanken:
I might be looking in the wrong way. I am running on Ec2 on docker - My app/media/project folder is completely empty - actually the folder does not exist.

I have found the old files here:
/var/lib/docker/volumes/517c3b5801359e483c12f516570af10c5978004c440ca511ca3af3d7915fc75a/_data/project/2/

and my new projects are here: 
/var/lib/docker/aufs/mnt/ff7a2fa1dd1657f7643e176191bdf4f4e5b747d29dbb7fddfe82ec9f5f79e059/var/www/data

pierotofy:
This is now solved in https://github.com/pierotofy/WebODM/commit/9b3f0843d0363cda92284a2e491883f9d40d70c4

I will fast-track the merge of this branch since it probably caused other users to see their data disappear. Note that the data is still in a docker volume and hasn't been lost, it just hasn't been mounted back after update.

pierotofy:
@hblanken you if update webodm the database will reset once more, but subsequent updates should maintain data persistently.

Please let me know if the problem happens again! Thanks for reporting the problem.

smathermather:
Ahh, this is nice... .

Issue #110, opened 2017-03-01
When tasks are moved in the database the filesystem is not updated


mbandy:
I have a working install for the use of my group and it is great. Thanks so much for this wonderful software!

I think this can be classified as a bug. When a task is reassigned to a different project in the administrative interface, the assets are not moved in the filesystem (app/media/project). So the assets (and source photos) are no longer available via the web interface. It's simple enough to move them by hand in the shell (after which everything works as expected) but WebODM should probably take care of this behind the scenes. 

Thanks again!

pierotofy:
Yes this is definitely a bug, along with #108. I will put priority on both of these in the next week or two.

Thank you for reporting the problem!

pierotofy:
This is fixed here https://github.com/pierotofy/WebODM/commit/0fa38e310bbb2f3bdeaef53e188f8f1441e77fb3, changes which I hope to land into master next week.

pierotofy:
Note that the branch still requires more testing, I wouldn't merge it just yet.

Issue #107, opened 2017-02-25
Processing node not recognized


mbandy:
Hi! Great project. I am very excited to use it. I am hoping someone can help me out.

I have a working native install of OpenDroneMap, WebODM and node-OpenDroneMap on Ubuntu 14.04. ODM works and has been tested on multiple datasets. Node-ODM also works. I can connect to it using a web browser, upload photos, and download assets. WebODM also seems to work, since I can log in, create/modify users, and so on.

Here's the probem. I can add the node-ODM instance as a processing node via WebODM and it adds sucessfully. However, if I make a new task and upload images I get the following error in a popup. I have added the proper permissions to the user/group and am at a loss as to how to proceed. Help would be much appreciated. Thanks!

> 

```
> There are no usable processing nodes. We tried to reach:
>        10.100.103.55:3000 (queue: 0)
> Make sure that at least one processing node is reachable and that you have granted the current user sufficient permissions to view the processing node (by going to Administration -- Processing Nodes -- Select Node -- Object Permissions -- Add User/Group and check CAN VIEW PROCESSING NODE). If you are bringing a node back online, it will take about 30 seconds for WebODM to recognize it.
```

pierotofy:
Hi @mbandy thanks for reporting the issue!

I'm guessing that the problem has to do with your network configuration; 10.100.103.55 is an address that must be reachable from the machine that is hosting WebODM.

| WebODM  |    ---------> |  10.100.103.55 |

The fact that you can access it independently:

| Your machine  |    ---------> |  10.100.103.55 |

Suggests that you might have to configure your firewall and/or network routes on the WebODM host.

If WebODM and node-OpenDroneMap are running on the same machine, simply use "localhost" for the IP address instead of 10.100.103.55.

mbandy:
HI, @pierotofy  thanks for your quick response!

I apologize, I should have included a bit more information in my question. I am pretty sure there is no network/routing issue. WebODM and node-ODM are running on the same machine. I have added the  processing node as localhost and also as 127.0.0.1 and WebODM exhibits the same behavior (below). And I can access 10.100.103.55:3000 from a shell environment (using telnet or a browser like links). So the machine can definitely connect to itself using any three of those network addresses. I am stuck!

```
There are no usable processing nodes. We tried to reach:
10.100.103.55:3000 (queue: 0)
localhost:3000 (queue: 0)
127.0.0.1:3000 (queue: 0)
[...]

Issue #80, opened 2017-01-24
Add Option to Download Textured Model Assets


pierotofy:
It would be nice to add the option to download all textured model assets in .zip format.

pierotofy:
This is now added in the `bugfixes` branch: https://github.com/OpenDroneMap/WebODM/commit/a907f2f7c27f4267d32d9a818a9c8eabad57d414

PR #116, opened 2017-03-08
Bugfixes

### LINKED ISSUE #108: A user can modify a Task's project_id, thus bypassing security contraints
Certain fields in a Task should not be editable, such as the project field, which allows a user to reassign a task to a project he/she does not own or has permission to access.

pierotofy:
Fixes #108  #115 #80 #110 #107.

@dakotabenjamin I plan to merge this by tomorrow, since #115 is quite critical. Let me know if you find anything that should be changed.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -419,15 +419,15 @@ class Task(models.Model):
             }
         }
 
-    def get_textured_mesh_archive(self):
+    def get_textured_model_archive(self):
         archive_path = self.assets_path("odm_texturing.zip")
-        textured_mesh_directory = self.assets_path("odm_texturing")
+        textured_model_directory = self.assets_path("odm_texturing")
 
-        if not os.path.exists(textured_mesh_directory):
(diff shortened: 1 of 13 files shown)


app/models.py
```diff
@@ -419,6 +419,18 @@ def get_tile_json_data(self):
             }
         }
 
+    def get_textured_mesh_archive(self):
+        archive_path = self.assets_path("odm_texturing.zip")
+        textured_mesh_directory = self.assets_path("odm_texturing")
```
dakotabenjamin: I don't know what self.assets_path refers to, but you might not want to hard-code the name here

pierotofy: Yes, I've now put it into a separate method "get_textured_mesh_filename".

dakotabenjamin: :+1:

pierotofy:
Thanks @dakotabenjamin !

PR #120, opened 2017-03-15
Devenv

pierotofy:
This PR aims to facilitate the setup of a development environment and help people get hacking on WebODM using docker.

@mojodna would you be able to take a look when you have a moment?

### DIFF
--- a/devenv.sh
+++ b/devenv.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+set -eo pipefail
+
+./webodm.sh checkenv
+
+usage(){
+  echo "Usage: $0 <command> [options]"
+  echo
+  echo "This program helps to setup a development environment for WebODM using docker."
+  echo 
+  echo "Command list:"
+  echo "	start		Start the development environment"
+  echo "	stop		Stop the development environment"
+  exit
+}
+
+
+run(){
+	echo $1
+	eval $1
+}
+
+start(){
+	run "docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml -f docker-compose.dev.yml up"
+}
+
+stop(){
+	./webodm.sh stop
+}
+
+if [[ $1 = "start" ]]; then
+	echo "Starting development environment..."
+	start
+else
+	usage
+fi
--- a/devenv.sh
+++ b/devenv.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -eo pipefail
+__dirname=$(cd $(dirname "$0"); pwd -P)
 
-./webodm.sh checkenv
+${__dirname}/webodm.sh checkenv
 
 usage(){
   echo "Usage: $0 <command> [options]"
@@ -25,7 +26,7 @@ start(){
 }
 
 stop(){
-	run "./webodm.sh stop"
+	run "${__dirname}/webodm.sh stop"
 }
 
 runtests(){
--- a/devenv.sh
+++ b/devenv.sh
@@ -9,11 +9,11 @@ usage(){
   echo "This program helps to setup a development environment for WebODM using docker."
   echo 
   echo "Command list:"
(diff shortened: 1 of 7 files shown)


devenv.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -eo pipefail
+
+./webodm.sh checkenv
```
mojodna: This will only work if `$PWD` is the directory containing `devenv.sh`. An idiom to determine the directory containing the current file is:

```bash
__dirname=$(cd $(dirname "$0"); pwd -P)
```

So you can then reference `webodm.sh` as `${__dirname}/webodm.sh`

docker-compose.dev.yml
```diff
@@ -0,0 +1,6 @@
+version: '2'
+services:
+  webapp:
+    entrypoint: /bin/bash -c "chmod +x /webodm/*.sh && /webodm/wait-for-it.sh db:5432 -- /webodm/start.sh --create-default-pnode --setup-devenv"
```
mojodna: Is it possible to skip the `chmod` by ensuring that files have appropriate permissions in the git repo?

slate/source/includes/_fordevelopers.md
```diff
@@ -0,0 +1,76 @@
...
+
+That's it! You can modify any of the files, including SASS and React.js files. Changes will be reflected in the running WebODM instance automatically. 
+
+### Native Setup
+
+If you can follow the instructions to [run WebODM natively](https://github.com/OpenDroneMap/WebODM#run-it-natively), you should be able to make changes to the code directly.
+
+## Run Unit Tests
+
+We believe testing is a necessary part of deliverying software of a certain quality. We try to achieve complete test coverage for backend code, while we are more relaxed about front end testing. See [#119](https://github.com/OpenDroneMap/WebODM/issues/119).
```
mojodna: "delivering"

start.sh
```diff
@@ -34,8 +34,25 @@ if [ $? -ne 0 ]; then
     exit
 fi
 
-echo Building asssets...
-webpack
+if [ $1 = "--setup-devenv" ] || [ $2 = "--setup-devenv" ]; then
```
mojodna: You probably mean `==`.

start.sh
```diff
@@ -34,8 +34,25 @@ if [ $? -ne 0 ]; then
...
+    echo Setup git modules...
+
+    git submodule init 
+    git submodule update
```
mojodna: These can be folded into `git submodule update --init`

start.sh
```diff
@@ -34,8 +34,25 @@ if [ $? -ne 0 ]; then
...
+    
+    echo Setup npm dependencies...
```
mojodna: Might be worth using the `__dirname` idiom above to ensure that you're starting in the right directory.

devenv.sh
pierotofy: Nice! I will add that!

docker-compose.dev.yml
pierotofy: I added that extra chmod command because people on Windows were having issues when cloning the repository.

start.sh
pierotofy: I didn't know this, but `=` and `==` are aliases. http://tldp.org/LDP/abs/html/comparison-ops.html

From testing it seems to be so.

start.sh
pierotofy: Did not know that, thanks!

start.sh
pierotofy: Good idea, usually this file is called from Docker which is always guaranteed to start from the proper directory, but perhaps other people might want to use this script natively.

pierotofy:
Thanks a lot for your feedback @mojodna ! :tada:

PR #123, opened 2017-03-19
Removed elevation default from unreferenced model view (replaced with…

pierotofy:
Show RGB by default even on unreferenced datasets


pierotofy:
Merging right away as it's a minor change

PR #125, opened 2017-03-21
Added CORS module to Django

pierotofy:
This PR adds CORS headers to Django. Needed for accessing the API via Web.


smathermather:
👍

PR #128, opened 2017-03-23
Upgraded Potree to 1.5RC

pierotofy:
This PR replaces the old version of potree with 1.5RC. This much improved version adds measurements, height profiles, more shading options, speed improvements and lots more.

image

Please let me know if you find any issues.


dakotabenjamin:
Playing around with it now. :+1: this is awesome I was just thinking we need to be able to do other navigation types. Thanks for reading my mind :laughing: 

- it would be helpful if the type of navigation was highlighted when selected
- The tooltip for coordinate measurement says angle measurement
- Textured model could use navigation tools
- Volume doesn't show up in measurements tab
- Can't delete/remove volume measurements (box and clipping)
- Text color in About tab is black on dark grey

pierotofy:
- I'll defer updating this one since it touches potree Viewer API code. One thing to do in the future as part of a major update would be to rebuild the viewer interface using React.js, instead of using the built-in Potree viewer code.
- Fixed
- Fixed
- Potree bug, should be fixed here https://github.com/potree/potree
- Same as above
- Fixed

Very nice feedback! Thank you!

smathermather:
This is flipping amazing. I got to play with it today. One possible regression is that I can no longer display 3D data in Firefox on Windows (I haven't tried Mac OS yet).

PR #156, opened 2017-04-18
Add ability to run as a systemd service on linux servers easily.

pmumby:
I've added a linux systemd unit file, and instructions on it's installation/use to the readme. This is tested on Ubuntu 16.04 server, and works great to auto start/stop the service running within the docker container, rather than requiring native installation (with manual pre-req handling).


pierotofy:
Looks fantastic, thank you!

smathermather:
This is awesome. Thanks @pmumby.

PR #161, opened 2017-04-23
stop: remove-orphans

falu:
remove orphans at stop


pierotofy:
Looks great! Thank you @falu !

PR #170, opened 2017-05-15
Update Mobile UI and GCP list file acceptance 

gkconsulting:
I have updated the CSS so that it is bit more mobile friendly, however it still needs some testing to make sure the changes are stable across multiple web browsers. I have also update the acceptance files so that gcp_list.txt files can be added to a project. I have tested it with two data sets and it seems to handle it fine, not sure if this should be a separate upload button? Sorry if there are any errors, first time using Github.

### DIFF
--- a/app/static/app/css/main.scss
+++ b/app/static/app/css/main.scss
@@ -183,3 +183,47 @@ button i.glyphicon{
         margin-right: 4px;
     }
 }
+
+/* This is to make the Header and navigation more mobile friendly */
+
+.navbar-toggle {
+    position: relative;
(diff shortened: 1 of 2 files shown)


pierotofy:
Looks great! I was expecting the GCP to fail because the ImageUpload model in Django is well, defined as an ImageField, but apparently Django does not validate the type of file, so the file is happily passed along to node-OpenDroneMap.

We have longer-term plans to add a GPC interface to WebODM, but this is an excellent addition for the time being.

Thank you for the contribution!

PR #175, opened 2017-05-21
Gunicorn

### LINKED ISSUE #173: Replace Django's server with gunicorn (or other)
Right now people are running WebODM off of django's built-in server, which is not suited for production use.

This might be causing issues like #163, among security and speed issues.

We should switch to something like http://gunicorn.org

pierotofy:
This PR adds gunicorn + nginx support (which was long overdue). In the process I have upgraded a few packages to solve some minor issues and changed some areas of the codebase to work with the new upgrades due to breaking changes.

@mojodna I know you've used gunicorn before, does this look all right?

I plan to merge these changes by Wednesday. Fixes #173


smathermather:
w00t!

PR #180, opened 2017-05-24
change permissions on init-db.sql to prevent permissions error on build

mspringfield:
Permission change to init-db.sql after copy to prevent this error: https://gist.github.com/mspringfield/d72a61d20149d43088f6d26425072500


pierotofy:
Looks great, thank you @mspringfield !

PR #250, opened 2017-07-25
Badinterfix

pierotofy:


pierotofy:
Fixes the infamous bin/bash bad interpreter error in some docker-compose configurations.

PR #271, opened 2017-08-24
Solve some random typos

patrickelectric:


app/static/app/potree/workers/BinaryDecoderWorker.js
```diff
@@ -270,7 +270,7 @@ Potree.Version.prototype.equalOrHigher = function(version){
 	}
 };
 
-Potree.Version.prototype.upTo = function(version){
+Potree.Version.prototype.up to = function(version){
```
pierotofy: This is actually a syntax error and will generate an error. We should reverse this change.

patrickelectric: Ooops

patrickelectric:
@pierotofy done :)

pierotofy:
Fantastic, thank you!

PR #295, opened 2017-09-29
Frontend tests

pierotofy:
This PR will add basic smoke test for the front-end code. This is a basic starting point, but at least we'll know if a change breaks javascript compilation and React components basic initialization and setup.


pierotofy:
Plan to merge these by tomorrow EOD. Comments are welcome!

Issue #352, opened 2017-12-01
Multiple File Selection Step


pierotofy:
People should be allowed to select pictures in chunks (add pictures from one folder, then add pictures from another folder, then drag some more, etc.).

Issue #310, opened 2017-10-30
Integrate --rerun parameter


tobi-kellner:
[This is more of a feature request, apologies if this is the wrong place for them, but there seem to be a few of these here]

In "naked" ODM I can re-run individual processing tasks using the **--rerun parameter**.
Using WebODM, I cannot seem to do that. 

This is unfortunate in my scenario as I am battling with crashes, mostlikely due to insufficient RAM, in the OdmMeshing process - around 7h into the process. I would like to experiment with different settings to see how they influence behaviour during this task but it is a pain if I have to run everything from the beginning every time.
Similarly, experimenting with different DEM/DSM parameters is easier if you don't have to run everything from step 0.

In the longer run, what would be awesome would be a bit more info on the order and dependencies of taks/steps, including which parameters influence what.
It's probably obvious to you all, but for noobs like me it is not always clear which tasks build upon each other, and which parameters have effects "downstream".
Provided the raw information, I would happily help writing this up or turning it into a table or diagram.

Thank you for your amazing work!

Greetings from Freiburg, Germany,
Tobi

pierotofy:
Hey Tobi, thanks for the feedback. This feature has been on the radar for a while, see #229. I think this will probably be available sometimes early next year.

tobi-kellner:
@pierotofy, thank you! 
And apologies for the double-post. Reminds me to always search first...

Btw, from a newbie perspective, it is a bit confusing what is currently the best place for raising issues and asking questions. 
There is this forum, then there is
http://community.opendronemap.org/c/webodm
and then there is the chat at
https://gitter.im/OpenDroneMap/generalhelp

I want to be good and play by the rules, and not ask stupid questions in the wrong places, but it's not so easy to figure out what these are :)

Thanks a lot for your help,

Tobi

pierotofy:
No worries, we're in the process of phasing out the gitter chat and slowly pushing the forum at community.opendronemap.org.

Issues (bugs, faults) go here (Github)
Everything else goes on community.opendronemap.org/c/webodm

pierotofy:
This has now been implemented as part of #380. Thanks again for the feature request!

Issue #316, opened 2017-11-07
WebODM AR Mobile App


pierotofy:
Create a iOS or Android (or both) app able to:
 - Log-in to a WebODM instance
 - Pull list of processed tasks
 - Allow the user to select a textured 3D model
 - Select a flat area onto which to project the model
 - Allow the user to navigate the model in AR

 - Bonus: allow measurements, display point clouds as well.

gl3nda85:
Hey i would be interesting in working on this in my spare time, have you begun??

pierotofy:
Barely, I've started a react-native app skeleton at https://github.com/pierotofy/webodm-ar. I'm currently doing work on the Expo framework to add support for plane detection (it's currently missing from Expo THREE). Once that's added, the app can be built. In the meanwhile there are non-AR things that could be worked on, such as login screen, project list selection, etc.

Would love to collaborate if you're interested!

gl3nda85:
hey dude, sorry for the late reply, i actually have no knowledge of react and have been using vuejs, maybe we can email or talk over slack to work something out? cheers

pierotofy:
So https://github.com/expo/expo-three/pull/13 has been merged, which means work on this can go forward.

The reason for using React is not because of a choice of JS framework, it's because it's a React Native application, which means we can easily use three.js and Apple's ARKit (and someday soon possibly Android's ARCore).

To my knowledge there's no React Native equivalent for vuejs.

pierotofy:
I'm closing this since I have no bandwidth to implement it and priorities have changed.

PR #380, opened 2018-01-25
Ability to restart tasks using rerun-from, multi-step files selection

pierotofy:
This solves #352 and #310.

Do not merge yet, still need to add unit testing for new areas of code, document API changes.

Requires the latest node-opendronemap changes from master to be tested.

image


pierotofy:
Totally should have opened a different branch for this, but whatever, we have support for sharing maps with QR codes now. :cheers:

image

Should enable a good workflow for interacting with #316

PR #403, opened 2018-02-23
Volume measurement tool 

Abizem:
Volume measurement implementation within ODM


app/static/app/js/components/Map.jsx
```diff
@@ -7,6 +7,8 @@ import Leaflet from 'leaflet';
 import async from 'async';
 import 'leaflet-measure/dist/leaflet-measure.css';
 import 'leaflet-measure/dist/leaflet-measure';
+import 'leaflet-draw/dist/leaflet.draw.css';
+import 'leaflet-draw/dist/leaflet.draw';
```
pierotofy: You might need to add `leaflet-draw` in `package.json` as well!

app/static/app/js/components/Map.jsx
```diff
@@ -139,7 +141,7 @@ class Map extends React.Component {
                                     <i class="fa fa-cube"></i> 3D
                                 </button>`;
 
-            layer.bindPopup(popup);
+            //layer.bindPopup(popup);
```
pierotofy: Is this disabling the popup permanently?

app/static/app/js/components/Map.jsx
```diff
@@ -184,6 +186,78 @@ class Map extends React.Component {
...
+      e.layer.bindPopup('Volume: ' + paramList.toFixed(2) + 'Mètres Cubes (m3)');
+    });
+
+    this.map.on(L.Draw.Event.EDITED, function(e) {
+      e.layers.eachLayer(function(layer) {
+        var paramList = null;
+        $.ajax({
+            type: 'POST',
+            async: false,
+            url: '/api/projects/1/tasks/4/volume',
```
pierotofy: Project ID and task ID are hard-coded here, perhaps replace them with:

`
const meta = layer[Symbol.for("meta")];
meta.task.project --> project id 
meta.task.id --> task id
`

app/static/app/js/components/Map.jsx
```diff
@@ -184,6 +186,78 @@ class Map extends React.Component {
...
+      
+    this.map.on(L.Draw.Event.CREATED, function(e) {
+      e.layer.feature = {geometry: {type: 'Polygon'} }; 
+      featureGroup.addLayer(e.layer);
+
+      var paramList;
+        $.ajax({
+            type: 'POST',
+            async: false,
+            url: '/api/projects/4/tasks/7/volume',
```
pierotofy: Same (see below)

app/static/app/js/components/Map.jsx
Abizem: Yes, of course

app/static/app/js/components/Map.jsx
Abizem: Maybe, we could find a way to disable this popup when the user start drawing the polygon for volume estimation!

app/static/app/js/components/Map.jsx
Abizem: right! i wwould test this,
const meta = layer[Symbol.for("meta")]; 
meta.task.project=project.id 
meta.task.id=task.id

app/static/app/js/components/Map.jsx
pierotofy: I think that would be a good way to do it, yes.

pierotofy:
I merged these changes into the volume branch, where I will make some modifications to bring this feature to completion.

Thanks @Abizem!

PR #425, opened 2018-04-02
Delete Grass GIS Mask removal

kikislater:
Due to GRASS environment destroyed each time computation is done, mask removal is not needed here.


pierotofy:
Looks great, thanks @kikislater !

PR #432, opened 2018-04-15
Up the recommended minimum overlap + 3D rec

smathermather:
Recommended overlap should be a little higher-- 65% guarantees 8 images per location, which is a better practice. I also added a recommendation for great 3D which is 83% overlap.


pierotofy:
Great recommendations, thanks @smathermather ! :+1:

kikislater:
Nice recommandation. Could I ask why 83 and not 80 or 85 ? Is there a formula behind ?

PR #433, opened 2018-04-15
Add pip to requirements list

smathermather:
Pip is the one missing requirement from this setup.


pierotofy:
True!  `./webodm.sh` attempts to install it automatically using easy_install if it's missing, that's why I didn't list it directly, but wouldn't hurt to tell users that it's a requirement.

Thanks!

PR #437, opened 2018-04-17
Fix some things

Wlanfr3ak:
Added more apt-get install data and fix Filename for pip install. Git Clone is now included.


pierotofy:
Looks great, thanks @Wlanfr3ak!

PR #467, opened 2018-06-17
IdEditor

### LINKED ISSUE #462: Don't create "First Project" for all users
We should probably limit the creation of the first project only to the first login for the admin user.

pierotofy:
Makes the process of opening the IdEditor easier (no need for manual copy paste of the background URL anymore).

Also fixes #462


smathermather:
w00t!

PR #514, opened 2018-09-04
Extract basemap config

mojodna:
This allows basemaps to be less obtrusively customized at build time (i.e. without needing to touch `Map.jsx`).

Context: https://github.com/posm/posm-build/pull/114


pierotofy:
Good addition, thanks @mojodna!

Issue #511, opened 2018-09-03
Always show a button to download or copy to clipboard task outputs


pierotofy:
This can be useful for debugging.

aaj013:
I'd like to work on this!

pierotofy:
I don't think anyone is currently working on this, so go for it :+1:

aaj013:
Please check the buttons positioning in UI (screenshot below)
![copy-download-task-outputs](https://user-images.githubusercontent.com/5847582/47155341-23463880-d302-11e8-892f-a6421654a65b.png)

pierotofy:
PR merged. Thanks!!

pierotofy:
@aaj013 The current webodm docker image does not have the PR as of yet. We'll include it soon. No action required on your part.

PR #539, opened 2018-10-18
Always show a button to download or copy to clipboard task outputs

aaj013:
This PR is against the issue https://github.com/OpenDroneMap/WebODM/issues/511
* Added buttons to download and copy to clipboard task outputs
* Added copyTaskoutput function


pierotofy:
Looks great!! Thanks @aaj013!

Issue #569, opened 2018-11-30
Change references of opendronemap/node-opendronemap to opendronemap/nodeodm


pierotofy:
Per recent changes to docker image names.

Issue #568, opened 2018-11-30
Upload interface "lags" when uploading a large amount of images


pierotofy:
The progress bar lags behind and the interface becomes unresponsive when uploading a large amount of images. 

This is probably due to too many `setState` calls that update the progress values (I think the calls are linear to the number of files in the queue).

Should be fixed.

pierotofy:
Fixed.

Issue #565, opened 2018-11-29
Add support for SSL processing nodes


pierotofy:
This has been in the TODO list for a while. We could infer the use of SSL from the port choice (443).

Issue #564, opened 2018-11-28
Highlight offline processing nodes in menu


pierotofy:
This is a minor UI improvement, but it would be cool to know the online/offline status of a node by just expanding the menu, without clicking on a node.

pierotofy:
Done.

PR #571, opened 2018-12-04
Improvements

### LINKED ISSUE #562: If node-odm returns a JSON error while querying /info, WebODM does not handle the error gracefully
Problem is in `nodeodm/models.py:71`. We should check for a 'error' key and handle it.

pierotofy:
Cumulative fixes and improvements. Fixed #562


pierotofy:
Fixed #569

pierotofy:
Fixed #565

pierotofy:
image

#564

pierotofy:
Fixed #568.

PR #580, opened 2018-12-31
Lightning Plugin

pierotofy:
This PR adds a Lightning Network plugin which helps users keep in sync their account on webodm.net with WebODM. Additionally it provides automatic credits estimation during task creation (only when appropriate), processing node syncing, permissions management and balance display.

Credit estimation still work in progress, do not merge yet.


pierotofy:
Just needs some testing and should be good to go.

image

pierotofy:
image

PR #690, opened 2019-07-01
Adding ElevationMap plugin

nchamo:
Hey everyone!

I created a plugin to generate and preview an elevation map in the 2D view, and I thought it might be of use. Please don't hesitate to provide feedback or ask for changes. Also, if you don't want to include it into ODM, that's ok too :smiley: 

**Plugin:**
The new plugin allows users to preview and export an elevation maps based on the task's DEMs. It was based on the Contours plugin.
plugin-panel

* The interval setting can be either a number, or a string of the form: `5-10,20-30`. If it is a number, it will be used a step and the intervals will be autogenerated.
* The reference setting can be either 'Sea' or 'Ground' (the latter is only available if the task has a DTM generated). If 'Sea' is selected, then the intervals will be calculated based on the DSM. If 'Ground' is selected, then the intervals will be calculated between the difference from the DSM and the DTM.
* The noise filter setting configures the size of the kernel used for the [morphologic operation 'Close'](https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_imgproc/py_morphological_ops/py_morphological_ops.html) in OpenCV
* The projection is used as in the Contours plugin
* The elevation map can be previewed, or exported to the same formats as the Contours plugin.

One main difference between this plugin and the Contours one, is that instead of deleting the preview, you can hide it in from the Leaflet Layers control. I thought it made more sense to do so, and maybe do the same with the Contours plugin, so that you can control everything from one place.

Here is a small video capture of the plugin in action. it has a little bit of lag, but I think you can still understand what's going on.
elevation-map

One thing that may not be super clear in the gif, when you click on an interval layer, it shows the area it covers.

I think that's it!

Let me know what you think :smile:


app/static/app/js/components/Map.jsx
```diff
@@ -329,8 +333,21 @@ https://a.tile.openstreetmap.org/{z}/{x}/{y}.png
   }
 
   componentDidUpdate(prevProps) {
+    let opacity = this.props.opacity / 100;
+    
+    Object.values(this.autolayers._layers)
```
nchamo: The probably has to be a better way to do this, maybe by extending the autolayers control, but JS is not really my strengh.

pierotofy:
Hey @nchamo :hand: this is great!

A quick request: would it be possible to make two separate PRs (one for the dev mode changes and one for the plugin)? Reason is the dev mode changes can be merged quickly as-is, but I'll need to study some changes for the plugin. A few things that we should think about for the plugin:

 * An interface for plugins to declare additional pip requirements just like they can for node modules (we shouldn't touch the master requirements.txt or add new ones)
 * Moving the `react-select` component dependency into the packages.json of the plugin instead of changing the master packages.json
 * Create entrypoints to interact with the opacity slider instead of changing the component directly. Could you explain a bit more about the changes made there (purpose of the additional checkbox)?

nchamo:
Hey @pierotofy !

Just created a new [PR](https://github.com/OpenDroneMap/WebODM/pull/691) for the dev mode changes.

Regarding you comments:
- I deleted the dependency to `react-select`. Shouldn't have been there in the first place. It's related to some other changes I have on my fork.
- I'm not sure why the unit tests are failing. I just see a timeout after running `/code/run.sh`, which is weird, because I didn't modify anything related (that I know of). I'm removing `react-select` and dev changes to this branch, so they will run again and let's see what happens.
- The interface to declare aditional pip requirements, do you think it makes sense to add that logic to the same place where the plugins are built? In `app/plugins/functions`.
- I'm adding an aditional checkbox so that the user can determine if they want the base layer (for example the orthophoto) to be affected by the opacity slider. Right now, the opacity slider affects the orthophoto, but doesn't affect (for example) the contours generated by the contours plugin. Some of our users wanted the slider to affect the elevation map, but they also wanted to be able to contrast the elevation map against the orthophoto. So the checkbox allows them to do so. 
    - About the entrypoint so that plugins can interact with the slider by themselved, I'm not sure how to do that exactly (this is my first time working with JS/React). Any change you have some example of that somewhere else in the code?

pierotofy:
Thanks for splitting the PR! See if the changes I pushed there fix the travis tests here too.

I'm thinking that plugin writers could simply have a "requirements.txt" file in the root folder of the plugin directory, and if one exists, pip requirements are installed. This requires some thought as containers are immutable and any installed dependencies should be installed in a volume (probably in a folder of `app/media`). I'll explore some options in the upcoming days.

Makes sense for the checkbox. Do you think that a different approach would be to add the opacity slider directly within the ElevationMap panel? That way you wouldn't need to interact with the opacity slider.

As a side note, it's likely that I will move the opacity slider as a plugin in the near future (since orthophotos and DEMs can already be set to opaque by clicking them and using the slider from the popups). The main opacity slider is mostly there for convenience to set opacity on everything, but doesn't handle well edge cases like this one.

nchamo:
Hey @pierotofy !

Thank you for taking the time to go though the PR :smile: 

Regarding the opacity, I guess it makes sense to move all the logic to a slider in the plugin itself. That way we can avoid the checkbox and there is no need to register the plugin with the general slider. It sounds easier to do, and to use.

I'll undo those changes and create a slider in the plugin panel, makes sense?

And let me know what you come up with regarding the plugin requirements. Take into account that rasterio also required `libgdal-dev` to be installed (I also added it in the `Dockerfile`), so it's not only pip requirements we might need.

pierotofy:
I'm OK adding libgdal-dev directly into the Dockerfile for the time being, it's a single dependency and it's probably useful for other tasks in the future other than using rasterio. Sounds good about the changes! Thanks for taking the time to do that.

app/static/app/js/components/Map.jsx
```diff
@@ -263,6 +263,8 @@ https://a.tile.openstreetmap.org/{z}/{x}/{y}.png
       selectedOverlays: [],
       baseLayers: this.basemaps
     }).addTo(this.map);
+    
+    this.map.layerControl = this.autolayers;
```
nchamo: This is so that the plugin can add itself to the layer control. Probably can be done better, but I'm not sure how

pierotofy: Would a hook help? https://github.com/OpenDroneMap/WebODM/blob/master/app/static/app/js/classes/plugins/Map.js#L15 https://github.com/OpenDroneMap/WebODM/blob/master/app/static/app/js/components/Map.jsx#L209

nchamo: Done!

Thanks for pointing me in the right directon :smile:

pierotofy:
I'll probably have an implementation for the `requirements.txt` for individual plugins sometimes later next week. It should be feasible by having an "install" process when a plugin is activated that executes pip along with the --target flag and then add that directory into the PYTHONPATH of both the webapp and worker containers. https://stackoverflow.com/questions/47055613/pip-install-dependencies-in-custom-directory

nchamo:
Sounds great! 

Let me know when it's merged so I sync the branch and make the changes

pierotofy:
I made changes so that plugins can specify a `requirements.txt` file in the plugin directory. When enabled, the plugin will fetch and install the necessary dependencies and store them in a persistent directory under `app/media`. Packages installed there are independent of the rest of the system.

I've made some small changes to the plugin to work with the new changes. Due to the size of the extra Python packages (~250 MB) I think I will set the plugin to require manual activation after install from the "Plugins" panel, as enabling by default would increase the size of WebODM installs (hope that's OK).

Still needs some unit tests on the new additions to core (I will take care of those sometimes tomorrow).

nchamo:
Hey @pierotofy !

Sounds good to me, 250 MB is a lot to add by default.

And I'm glad you like it :smile: 
I'm more than happy to help!

pierotofy:
Merged! :clinking_glasses: 

Thanks @nchamo!

PR #691, opened 2019-07-01
Now plugins are watched in dev mode

nchamo:
When running WebODM in dev mode, plugins are also watched by webpack and so changes are reflected without having to rebuild


pierotofy:
Ok so it looks like the travis build was failing because WO_DEV was set to YES by default. I've changed it to be NO by default.

Thanks for the PR @nchamo !!

nchamo:
Thanks for the changes! :smiley:

PR #694, opened 2019-07-10
Cesium ion for WebODM

ProjectBarks:
### Overview
This is the version 1.0.0 release of the WebODM to ion plugin. Designed to take ODM assets and upload them to ion. This enables for future possibilities such as: viewing in CesiumJS, data fusion, analysis and integration into supporting platforms. 

### Screenshots
<p float="center">
  
   
  
</p>

### Features
- Uploading to Cesium ion
- Progress viewer 
- Direct viewing from tasks
- Naming based off project, task, and asset type

### Thoughts & Limitations
- While access to the lat, lng bounding box for the task is available there is no way to obtain its elevation for texture models which would help for localizing it to global terrain
- Currently there is no default Cesium viewer built in but I have one I am developing on the side for project level contextualization
- It would be super cool if we could do a blog post together!


pierotofy:
Hey @ProjectBarks :hand: this looks awesome! Thank you for the contribution. :pray:

I think a workaround to find elevation would be to manually parse the odm_textured_model_geo.obj Z coordinates as they are expressed in meters above sea level, although there might be a better way to do so, perhaps as a post processing step or as part of generating a report in ODM (perhaps by saving the output of `pdal info --summary odm_georeferenced_model.laz`).

```
"bounds":
    {
      "maxx": 576766.1718,
      "maxy": 5188210.142,
      "maxz": 168.6826152,
      "minx": 576665.3316,
      "miny": 5188115.55,
      "minz": 156.6415079
    },
```

I'd be happy to release a blog post.

PR #697, opened 2019-07-12
extending osm-quickedit [WIP]

danbjoseph:
Adjustments to the osm-quickedit plugin: 
- [x] Provide options for both iD editor and JOSM
- [ ]  ~Pick the topmost layer instead of the first on the list~

---------

Button is now a drop~down~up:
Screen Shot 2019-07-12 at 4 18 33 PM

Similar to the OSM website:
Screen Shot 2019-07-12 at 4 26 03 PM

Can load a background layer into iD editor:
Screen Shot 2019-07-12 at 4 17 17 PM

Can load an imagery layer into JOSM and download the OSM data extent:
Screen Shot 2019-07-12 at 4 27 56 PM

---------

Testing locally the imagery URLs return `404`. Clicking OSM Digitize from https://demo.webodm.org/dashboard/ (App Version: 0.8.1) doesn't show the processed imagery in the iD editor that opens and gives the same `404 (Not Found)`. Does it need to pass an [authentication token](https://docs.webodm.org/#orthophoto-tms-layer)? Should the API be extended so that it can issue "read only" tokens?


danbjoseph:
How to get the map to display multiple tasks? I have 2 projects, one of which has 2 tasks (different image sets). Viewing any of the 3 maps there is only ever one option in the Tasks section of the layer picker.
 
Screen Shot 2019-07-12 at 4 39 55 PM
Screen Shot 2019-07-12 at 4 40 07 PM

pierotofy:
To view multiple tasks, press the "View Map" button from the top right corner of the project. (example: https://demo.webodm.org/map/project/48/ ).

The 404 is a regression bug (nice find!) introduced by the time we updated Django to 2.1.

Fixed in #698. You will need to log-out and log-in once the changes are applied in your branch.

danbjoseph:
So, iD editor can load the tiles if you're in the same browser as you're logged into WebODM. But you can't load tiles in a different browser, or in a separate application like JOSM. 

Screen Shot 2019-07-15 at 12 27 08 PM

Screen Shot 2019-07-15 at 12 27 01 PM

pierotofy:
Well, right, that's expected; if you are not logged-in you don't have access to the tiles.

If you want to make the tiles available to anonymous users (from other browsers that are not logged-in), make sure you shared the task first.

danbjoseph:
Cool. It does work if you enable sharing of the task.

Screen Shot 2019-07-15 at 1 57 11 PM

Is there a better way to inform the user other than a `window.confirm()` on clicking to edit in JOSM? What's a good notification message?

Screen Shot 2019-07-15 at 1 59 40 PM

pierotofy:
I think a confirm would do fine. You could present a bootstrap modal if you don't like the UI of window.confirm, `FormDialog` is a JSX component to do that in WebODM, but you'll be looking at using React components. If you're not familiar with them, you might be in for some learning (`ShareDialog.jsx` is an example of using `FormDialog`).

pierotofy:
A different workflow could involve simply checking if the task is shared, and if not, automatically share it for the user before opening JOSM.

pierotofy:
Does this need more time @danbjoseph  ?

danbjoseph:
I got stuck on the "Pick the topmost layer instead of the first on the list"
and also want to make the imagery service URL available to copy from the share menu (to paste into QGIS or geojson.io or some place.
but expanding it to both iD editor and JOSM is done and seems to be good.

danbjoseph:
a map with more than one layer doesn't have the share option button so this means the same implementation for editing won't work?

pierotofy:
Right, the workflow for a map with multiple layers would require some more logic.

You let me know when this can be merged, I was just checking :pray: no rush!

JamesLeonDufour:
Hi same issue here, the tiles doesnt load in OSM even with shared activated.

JamesLeonDufour:
Need to review the export code to work in OSM
need to add /tiles/{zoom}/{x}/{y}.png  at the end of the tiles to make it works in OpenStreetMap, 

Can somebody adjust in the code. @pierotofy

pierotofy:
Updates on this @danbjoseph ?

danbjoseph:
i'm making a renewed attempt at this. i seem to remember just running `./webodm.sh start --dev` and being able to edit the plugin's `main.js` file and see the changes? but it's not happening. tried disabling/enabling the plugin. tried stopping and starting webodm. tried logging in and out. any idea what i'm doing wrong?

pierotofy:
It should still work the same (I've tried this with Linux/Firefox). Perhaps you need to refresh your browser's cache?

danbjoseph:
if a map has more than 1 task displayed, there is no share button and so i can't enable sharing for the editors? i've set the osm edit button to hide for that case. unless there's a way to check if a task is shared and enable it programmatically? right now i do it by checking the appearance of the share button and triggering a jquery click on it if i need to turn it on.

pierotofy:
Yeah don't worry about multiple tasks; if there's no share button, hiding is a good option. :+1:

danbjoseph:
Then I think this should be good to go for now.

plugins/osm-quickedit/public/main.js
```diff
@@ -1,26 +1,146 @@
...
+						});
+				}
+				
+				function prepareJOSM(){
+					var imageryParams = {
+		        title: tile.meta.name,
+		        type: "tms",
+		        url: encodeURIComponent("tms[22]:" + tileUrl)
+	        };
+					sendJOSMCmd('http://127.0.0.1:8111/imagery', imageryParams)
```
pierotofy, line 97: The JOSM button doesn't seem to work, I suspect because of this hard-coded address.

pierotofy: Rest looks great!

danbjoseph: Do you have JOSM open on your computer? Nothing will happen if it's not. That's the behavior for both https://www.openstreetmap.org/ and https://tasks.hotosm.org/

danbjoseph: I thought those websites used to warn you if there wasn't a response and would show a prompt asking you if JOSM was open. But I don't see that happening now.

pierotofy: Ahh, no I don't, what's why. Never mind then! Looks good!

pierotofy:
Merged. Thanks @danbjoseph :+1:

smathermather:
This is fantastic. I was using workarounds to accomplish this.

danbjoseph:
i'm not seeing the "Cannot communicate with JOSM." error you've added (when i click and JOSM isn't open). did it display on your computer? i used code from the HOT OSM Tasking Manager for sending commands to JOSM and so don't fully understand it (well, I don't fully understand Promises). but the `sendJOSMCmd` is mostly just to space things out? does JOSM indicate back if a command is received successfully and is it captured by the function?

pierotofy:
It displayed on my computer, so not sure.

danbjoseph:
Ah, works in Firefox. Not Chrome. 👍

PR #700, opened 2019-07-15
[cesium-ion] Add Texture localization + Fix delete refresh bug

ProjectBarks:
- Fixed texture localization so all "3D Texture" models re-uploaded should now be referenced to their geospatial position. This is in response to [this report](https://community.opendronemap.org/t/bringing-cesium-ion/2384/15).
- When an asset was being refreshed the user was not getting that UI update until a second refresh. This was due to a bad destructure. 
- Bumped version to 1.2.0


ProjectBarks:
@pierotofy looks like an unrelated issue caused the build to crash.

pierotofy:
Thanks @ProjectBarks ! Looks great. I've fixed the build issue in #701, so this is safe to merge.

PR #702, opened 2019-07-16
[cesium-ion] Texture model prune

ProjectBarks:
- Texture models duplicate their internal data so files are now trimmed based off the availability of geo-referenced files or it will fallback to default files
- Bumped version


pierotofy:
Thank you @ProjectBarks ! Great to see the plugin improve.

PR #704, opened 2019-07-17
Specify the user model using the AUTH_USER_MODEL setting

akai16:
Specify the user model using the AUTH_USER_MODEL setting as discussed in the Docs.
https://docs.djangoproject.com/en/dev/topics/auth/customizing/#referencing-the-user-model


pierotofy:
Great improvement, thank you @akai16 :pray:

PR #706, opened 2019-07-21
Fix console output off-by-one error

pierotofy:
The first line of the console output was always truncated due to an off-by-one error.


smathermather:
For the record, I created that issue label a long time ago...

PR #710, opened 2019-07-31
Fixed spelling error "Sympthoms" -> "Symptoms"

octopuscabbage:
I'm actually not really sure symptoms is the best word there.


pierotofy:
Ah, nice catch! Thank you! :+1: :pray:

PR #712, opened 2019-08-02
Get settings.py through django.conf

akai16:
The previous "from webodm import settings"  would get the first settings.py
Using django.conf is a better option 
This is discussed in this Stack Overflow topic: https://stackoverflow.com/questions/19976115/whats-the-difference-between-from-django-conf-import-settings-and-import-set
And in the docs: https://docs.djangoproject.com/en/dev/topics/settings/#using-settings-in-python-code


pierotofy:
Thanks @akai16 :+1: looks great.

PR #716, opened 2019-08-13
[Up for discussion] Running webodm in detached mode

nchamo:
Hey ODM maintainers!

I've been having a problem when I start WebODM. If the container doesn't exist, `docker-compose up` creates it, starts it and attaches by default. This causes a few problems when I want to do something else after WebODM starts, since the terminal is blocked.

Also, what is kind of weird, is that if the container is only stopped, `docker-compose start` doesn't attach. So the blocking depends on the state of the container.

My proposal is to only attach if WebODM is run in **debug** or **dev** mode. This would change the current behavior of how WebODM starts. 

If you would like to avoid that, another option could be to add a new flag (maybe `--detached`) that always runs in detached mode. That wouldn't change the current funcionality.

A third option would be to close the PR and run WebODM in the background in the first place, but I thought it might make sense to bring all this up.

What do you think? :smile:

### DIFF
--- a/webodm.sh
+++ b/webodm.sh
@@ -20,6 +20,7 @@ fi
 
 load_default_node=true
 dev_mode=false
+debug_mode=false
 
 # Load default values
 source .env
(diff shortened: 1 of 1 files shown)


pierotofy:
I think there's certainly space for having a detached mode; I'd say it would be better not to change existing behaviors as most users (non-devs) would expect some sort of output coming from the `./webodm.sh start` command (with CTRL+C being the most intuitive way to stop the application).

Would it be possible to refactor this to have a `--detached` flag instead? :+1:

nchamo:
@pierotofy flag added :smile:

pierotofy:
Merged. Thanks so much @nchamo !

PR #718, opened 2019-08-14
Load multiple NodeODM nodes on WebODM startup

nchamo:
This PR gives non-tech users the ability to load multiple NodeODM nodes when starting WebODM.

In particular, I started using this when a node I had sometimes went offline (maybe related to this [post](https://community.opendronemap.org/t/localhost-3000-seems-to-be-offline/932/14)).

Let me know what you think :smiley:


app/migrations/0029_rename_default_odm_node.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from nodeodm.models import ProcessingNode
+
+def rename_default_node(apps, schema_editor):
```
nchamo: Docker names the nodes `webodm_node-odm_X`. So I'm migrating previously default nodes to the new name.

pierotofy:
Very cool addition! I wonder if this helps with #692 also. +1 for thinking of backward compatibility with older installations.

I've made a few small tweaks and used the `label` property to keep the node-odm names more readable in the UI.

Thanks for the contribution! :pray: I'll merge as soon the tests pass.

nchamo:
Glad to contribute to this awesome project :)

PR #724, opened 2019-09-08
Fix Cloud Import: Task Button Is Never Shown

nchamo:
Hey @pierotofy!

So in the PR we merged yesterday (https://github.com/OpenDroneMap/WebODM/pull/721), I added a little bug in one of the final changes. 

I basically wasn't returning the react element at any time, so the task button was never shown.

Apparently I hadn't rebuilt WebODM, or it wasn't on dev mode, but it looked like it worked, when it didn't.

I caught the problem today, when I synced my fork with the latest master 😄

### DIFF
--- a/plugins/cloudimport/manifest.json
+++ b/plugins/cloudimport/manifest.json
@@ -2,7 +2,7 @@
 	"name": "Cloud Import",
 	"webodmMinVersion": "1.1.2",
 	"description": "Import images from external sources directly",
-	"version": "1.0.0",
+	"version": "1.0.1",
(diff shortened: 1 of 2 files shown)


pierotofy:
:+1: ups, I can't believe I didn't catch it.

The use of `async: false` is OK, I wonder if a more elegant solution for the future could be to return a promise or use a callback (`async: false` will block, thus not allowing other code to execute asynchronously).

nchamo:
+1 on the promise or callback!

PR #728, opened 2019-09-13
Add admin api

w-toguchi83:
Hi.

I needed an API to add users.
So I implemented it this time.

`/api/admin/users/`
`/api/admin/groups/`

These APIs can only be used by administratos.

NOTE:
I implemented Django(and rest_framework) for the first time.
Therefore, it may not be good.

Please merge if you like.


pierotofy:
Hey @w-toguchi83 :hand: thanks so much for the PR! 

Would you be also interested in documenting the API by adding a new file in https://github.com/OpenDroneMap/WebODM/tree/master/slate/source/includes/reference so that it shows up in https://docs.webodm.org ?

w-toguchi83:
@pierotofy 

Yes. I will try to add a document.
But, I don't have much work time because it is a weekend.

Please take a few days.

pierotofy:
Sounds good! In the meanwhile I'll go ahead and merge this PR. You can open a new one for the docs when you have time :+1:

PR #729, opened 2019-09-18
Changed Ruby version of slate. Because gem didn't work.

w-toguchi83:
@pierotofy 

Hi.

The Slate didn't work in my local machine ( I chose to use vagrant. ).

I got errors.

```
ERROR:  While executing gem ... (NameError)
    uninitialized constant Gem::SafeYAML
```

or

```
YAML safe loading is not available. Please upgrade psych to a version that supports safe loading (>= 2.0).
ERROR:  Error installing bundler:
bundler requires Ruby version >= 2.3.0.
```

So, I fixed it that changed `Vagrantfile`. (Ruby version 2.0 to 2.4.7)

I confirmed that it works by accessing `http://localhost:4567` in my environment.

Check it, and please merge if you like.

### DIFF
--- a/slate/Vagrantfile
+++ b/slate/Vagrantfile
@@ -4,11 +4,21 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision "bootstrap",
     type: "shell",
+    privileged: false,
     inline: <<-SHELL
       sudo apt-get update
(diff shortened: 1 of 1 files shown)


pierotofy:
Perfect, thank you! :clap:

PR #759, opened 2019-12-04
[docker] fix #758 (containers not restarting on host reboot)

### LINKED ISSUE #758: Docker : containers not restarting after host reboot

Ubuntu 16.04

### What is the problem?

Containers do not restart after reboot (actually, here, webapp and worker do restart (unsure why?)  but not the others).

Containers should restart after reboot, so that the stack survives a host reboot without manual intervention. This should be an easy fix (just change `restart-policy` to `unless-stopped`

Run the stack with `./webodm.sh start -d`. See that it works. Reboot the host. See that it doesn't work.

olivierdalang:
Fix for #758 where container didn't reboot.

From docker-compose docs :
> unless-stopped : Similar to always, except that when the container is stopped (manually or otherwise), it is not restarted even after Docker daemon restarts.

I also removed the `:10` (that I think is supposed to limit the number of retries), as it is not documented in the docker-compose reference, and not necessarily a good idea (it may prevent the stack from recovering from transient issues such as connectivity issues)

Cheers !

### DIFF
--- a/docker-compose.nodemicmac.yml
+++ b/docker-compose.nodemicmac.yml
@@ -14,5 +14,5 @@ services:
     container_name: node-micmac-1
     ports:
       - "3000"
-    restart: on-failure:10
+    restart: unless-stopped
     oom_score_adj: 500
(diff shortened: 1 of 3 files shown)


pierotofy:
Looks great! :+1: 

Thank you @olivierdalang !

Issue #766, opened 2019-12-10
Volume Calculation is too large and Ortho image pixelated


dmbworld:
### What's your browser and operating system?
Firefox 71 on Windows 10
WebODM server running on Ubuntu 19.04
whatismybrowser.com/w/DMLG6JE

### What is the problem?
Volume calculation in current version is inaccurate and on average four to five times greater than in older version of WebODM which is actually the correct calculation. Additionally in the newer vesion of WebODM Ortho view the image is pixelated and scaled down.

### What should be the expected behavior?
Correct volume expected and not so pixelated image

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? Please include multiple screenshots of the problem! If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

Older WebODM Image with volume calculation:
![Older_WebODM](https://user-images.githubusercontent.com/58004227/70513038-8368f100-1b39-11ea-98a3-5322d41fc88a.png)

Newer WebODM Image with volume calculation:
![Newer_WebODM](https://user-images.githubusercontent.com/58004227/70513080-97acee00-1b39-11ea-99c9-550a817b25a1.png)

As can be seen in the newer version the image is pixelated, smaller in file size and shows an inflated volume of ariund four times the actual size in the older version.

pierotofy:
Thanks for the report @dmbworld 

I think this is a bug due to the fact that the DSM/DTM is now in web mercator projection (EPSG:3857), before we used UTM, and GRASS is now calculating values in that projection, which is expected to be off by:

```(1 / cos (latitude)) ^ 2`

For your area I think that's a factor of `5.20`, which explains the different results.

Still not sure about the lower resolution, but can confirm that results in the web viewer look more pixelated compared to opening them in QGIS.

dmbworld:
Just seen the merge and downloaded what appeared to be an update, calculation is still exhibiting the similar multiple too great.

pierotofy:
Docker images haven't been updated yet; try again in a few hours https://hub.docker.com/repository/docker/opendronemap/webodm_webapp

dmbworld:
Ok, thanks @pierotofy

dmbworld:
Yep, it has now updated correctly, sorry about preempting the update. Volumes now look as they should.
[...]

PR #767, opened 2019-12-10
Adjust volume measurements for web mercator

pierotofy:
Should fix volume measurements issues reported in #766


kikislater:
res=0.01 should slow down process : lot of computation !

pierotofy:
What would be a reasonable value? Leaving the parameter out and let GRASS handle the resolution choice?

PR #769, opened 2019-12-10
Fix histogram deformations

pierotofy:
Fixes https://community.opendronemap.org/t/possible-bug-weird-histograms-for-orthophoto/3248

### DIFF
--- a/app/api/formulas.py
+++ b/app/api/formulas.py
@@ -58,11 +58,13 @@ algos = {
     },
     'LAI': {
         'expr': '3.618 * (2.5 * (N - R) / (N + 6*R - 7.5*B + 1)) * 0.118',
-        'help': 'Leaf Area Index estimates foliage areas and predicts crop yields.'
+        'help': 'Leaf Area Index estimates foliage areas and predicts crop yields.',
+        'range': (-1, 1)
     },
     'EVI': {
         'expr': '2.5 * (N - R) / (N + 6*R - 7.5*B + 1)',
-        'help': 'Enhanced Vegetation Index is useful in areas where NDVI might saturate, by using blue wavelengths to correct soil signals.'
+        'help': 'Enhanced Vegetation Index is useful in areas where NDVI might saturate, by using blue wavelengths to correct soil signals.',
+        'range': (-1, 1)
     },
 
     # more?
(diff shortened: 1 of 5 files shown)


pierotofy:
image

pierotofy:
cc @smathermather

pierotofy:
Also better hillshade blending and fixed #770

dmbworld:
Here is a comparison of the OLD with the NEW:
OLD:
SM_Old

NEW:
SM_New

There is a difference in elevation delineation and there are still some lines/blocks visible in the New version.

pierotofy:
You can change the delineation by changing the `Shading` option to `Extruded`.

Can you annotate and/or zoom the lines/blocks you are seeing? Perhaps these are just hidden by the stronger shading in the previous version and are inherent to the data.

dmbworld:
The settings were Jet/Extruded:
SM_New_Jet_Extruded-2

And the Old one zoomed in:
SM_Old-2


Hope this helps.

smathermather:
Looks great!

dmbworld:
@pierotofy do you require any more details to help resolve the block/line visibilty feature.

pierotofy:
Did #771 fix it?

dmbworld:
Sorry @pierotofy, missed that update, only seen #771 just now, updated and this is the result:
SM_New_Jet_Extruded-3

Unfortunately the lines/blocks have become more distinct.

dmbworld:
@pierotofy Update: Unfortunately volume calculation is now broken. "Volume: computing... "and the cog is rotating with no value returned.
One of the CPUs rockets up to 100%.

dmbworld:
@pierotofy I've spotted #773, does this actually deal with the issue I've described concerning volume calculation speed?

dmbworld:
@pierotofy Thanks! #773 solves my mutterings above concerning volume calculation speeds.

dmbworld:
I took a different approach to identifying the reason for the visible blocks. Using the same dataset from April this year and the most current release of OpenDroneMapper and the default settings in both cases. Two tasks were created. One Task used Fullsize images and the other Resized. Viewing in the same system the results were:
FULL SIZE:
SM_Fullsize

RESIZED:
SM_Resized

The blocks are clearly visible in the resized version and this is not the case in earlier ODM versions where I always Resized.

Issue #777, opened 2019-12-17
NVDI and ENDVI from NGB cameras


Miguelistan:
Hi guys,

After the last improvements on ODM for crops, I noticed something very important is missing.
Since the NGB format is the most popular format for cameras for agricultural use, all the indexes obtained from them are vital for people working with drones in crops. 

Sadly there is no possibility today of getting NDVI from NGB cameras in ODM. I tried to add it myself, but since I am not a developer neither familiar with coding, I failed.  Here you can see the thread:

https://community.opendronemap.org/t/adding-ndvi-and-endvi-for-ngb-cameras/3290/5

Is there anybody that can help out with this? We farmers will thank you a lot.

pierotofy:
Hi @Miguelistan,

It looks like there is some information missing from your issue that will be needed in order to diagnose and fix the problem at hand. Please take a look at the [Issue Template](https://github.com/OpenDroneMap/WebODM/blob/master/ISSUE_TEMPLATE.md), which will tell you **exactly** what your issue has to contain in order to be processable.

Also, double check that this is the right place. If you are just asking for information, reporting feedback or proposing a few feature, the right place to ask is the [Community Forum](http://community.opendronemap.org/c/webodm), not here.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next week (until 2019-12-20 12:40) I'll close this issue so it doesn't clutter the bug tracker.

Cheers!
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

Miguelistan:
What's your browser and operating system?
Chrome and Windows

What is the problem?
It is not possible to process NDVI neither ENDVI with NGB cameras (most popular format for agricultural cameras)

What should be the expected behavior?
ODM should be able to process both indexes by using the following formulas:
- NDVI is (N-B)/(N+B)
- ENDVI is ((N+G) - (2B)) / ((N+G) + (2B))

How can we reproduce this?
You can follow this thread: 
https://community.opendronemap.org/t/adding-ndvi-and-endvi-for-ngb-cameras/3290/5

pierotofy:
Interesting, OK for ENDVI: `((NIR + Green) - (2 x Blue)) / ((NIR + Green) + (2 x Blue))`

NDVI technically uses the `Red` band, is this a variation (we can call it `NDVI (Blue)`)?

Miguelistan:
[...]

PR #778, opened 2019-12-18
Added ENDVI, NDVI (Blue)

pierotofy:
Adds the algorithms proposed in #777


philipperemy:
@pierotofy I've just started to use WebODM. Great project. Can you point me on how to use the algos/formulas? I'd like to convert some images to NDVI. Thanks!

Issue #649, opened 2019-04-04
Make volume measurements async


pierotofy:
Currently the calculation needs to return within the browser timeout limits, instead of pooling for status updates (like the contours plugin does).

ChipwizBen:
Are these able to be multithreaded also?

pierotofy:
Maybe, but not easily. GRASS GIS functions are mostly single threaded (I think that includes the ones we use for volume measurements). One would have to split the polygon area by the number of threads available and run the sum of volumes independently and add up the results.

But we would welcome contributions!

pierotofy:
This is fixed.

PR #782, opened 2019-12-20
Support for orthophotos with more than 3 bands

pierotofy:
Necessary changes required to display orthophotos with more than 3/4 bands.

Also new filters for proper band ordering.


pierotofy:
Fixed #790

pierotofy:
Adds #789

pierotofy:
Adds #649

pierotofy:
Note to self, this still needs unit tests.

Issue #756, opened 2019-12-01
SSL Certificate Install Error


ryan-shea:
### What's your browser and operating system? 

Firefox 70 on macOS (Catalina)

Ubuntu 18.04 TLS on GCE

### What is the problem?

When installing a new version of WebODM on GCE, I ran into a few errors with implementing SSL.

First, the autogeneration of certificates no longer works. The command in letsencrypt-autogen.sh includes --tls-sni-01, which is deprecated here: https://community.letsencrypt.org/t/how-to-stop-using-tls-sni-01-with-certbot/83210

When using the default --ssl flag, WebODM is then inaccessible.

I generated the keys myself with Certbot and passed them through to WebODM, only to run into another port allocation problem. On restart, WebODM allocates port 80 to a default nginx install, so the rest of the startup fails.

The only solution I’ve found is to pass a different SSL_INSECURE_PORT_REDIRECT. Then, WebODM works on port 443.

### What should be the expected behavior?

SSL Autogeneration succeeds + port 80 is unallocated on manual SSL

### How can we reproduce this?

Install any WebODM version and attempt to use SSL (Certbot is the issue here)

```
Will enable SSL (Manual)
docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml -f docker-compose.ssl.yml -f docker-compose.ssl-m
anual.yml start || docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml -f docker-compose.ssl.yml -f d
ocker-compose.ssl-manual.yml up --scale node-odm=1
Starting db       ... done
Starting node-odm ... done
Starting broker   ... done
Starting worker   ... done
Starting webapp   ... done
ERROR: No containers to start
Creating network "webodm_default" with the default driver
Creating db ... 
Creating webodm_node-odm_1 ... 
Creating broker ... 
Creating db
Creating broker
Creating broker ... done
Creating worker ... 
Creating worker ... done
Creating webapp ... 
Creating webapp ... error
ERROR: for webapp  Cannot start service webapp: driver failed programming external connectivity on endpoint webapp (
c40d60930f1a6dd39b78d181681188803ff7a21b7d0e98cf3d3861d5cfb06147): Error starting userland proxy: listen tcp 0.0.0.0
:80: bind: address already in use
ERROR: for webapp  Cannot start service webapp: driver failed programming external connectivity on endpoint webapp (
[...]

PR #785, opened 2019-12-31
Update letsencrypt to match current protocol

smathermather:


smathermather:
I don't know if this is what #756 referred to, but did my best to make the change... .

pierotofy:
Thanks @smathermather ! This should do.

SchulteMarkus:
For other guys ending here, not knowing what to replace ```--tls-sni-01-port```.

Use now:

```bash
--http-01-port
```

PR #793, opened 2020-01-21
Add multispectral preset

pierotofy:
Adds multispectral preset.


pierotofy:
Also proper rendering of N-band orthophotos:

image

dmbworld:
@pierotofy Just downloaded the most recent Docker images and all is now well. WebODM is running!
Many thanks for the super quick attention, I must have been building my error report at the same time as @ppicazo

PR #805, opened 2020-01-30
fix file upload in selinux enabled environments

FlorianMickler:
docker-compose.yml:
	 :z on volume mounts relabels the directory to enable access for all docker container
	 :Z on volume mounts relabels the directory to enable access for this specific docker container
	Some background:
		https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/
		https://docs.docker.com/release-notes/docker-compose/ (-> Syntax introduced in v1.4.0 from 2015-08-04)

	Since these volumes are designed only for use by webodm a relabeling to allow docker containers to have access seems resonable.

webodm/settings.py:
	The change in webodm/settings.py is necessary, because else the webapp will create big upload files in /tmp and move them to the final destination afterwards. Which means they will have container specific MCS context selinux labels applied to them. If the tmp files are created under MEDIA_ROOT, which has the :z labels applied to it, the files will get a label that is accessible from all docker containers.

	Some background:
		https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/selinux_users_and_administrators_guide/sect-security-enhanced_linux-working_with_selinux-selinux_contexts_labeling_files

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     ports:
       - "5432"
     volumes:
-      - dbdata:/var/lib/postgresql/data
+      - dbdata:/var/lib/postgresql/data:Z
     restart: unless-stopped
     oom_score_adj: -100
   webapp:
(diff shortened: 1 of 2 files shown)


pierotofy:
Hey @FlorianMickler :hand: thanks for the patch! Looks great.

PR #834, opened 2020-03-13
Fix #833

### LINKED ISSUE #833: Volume Measurement reports "Invalid task output"
### What's your browser and operating system?
whatismybrowser.com/w/ADN7Z5H

### What is the problem?

Volume measurement reports  "Invalid task output (cannot find valid key)"

Should report volume of area defined
Annotation 2020-03-13 234528

### How can we reproduce this? 
Attempt a volume measurement

pierotofy:
Fixes #833

### DIFF
--- a/plugins/measure/calc_volume.py
+++ b/plugins/measure/calc_volume.py
@@ -35,7 +35,7 @@ def main():
     Module("r.external", input=opts['dsm_file'], output="dsm", overwrite=True)
 
     # Set Grass region to vector bbox and resolution to DSM
-    Module("g.region", vector="region", res="dsm")
+    Module("g.region", vector="region")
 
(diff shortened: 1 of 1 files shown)


kikislater:
Sorry my mistake ! Didn't check and wrong parameter

I should be : 
```
     Module("g.region", raster="dsm") 
     Module("g.region", vector="region") 
```

instead of 
```
     Module("g.region", vector="region", res="dsm") 
```

It's the easiest way (set region to raster resolution and extent, then set extent to vector layer), otherwise, it needs to extract resolution from r.info

dmbworld:
Thanks @pierotofy, checked and works.

pierotofy:
No worries, feel free to open a new PR @kikislater !

PR #835, opened 2020-03-15
Allow accurate measurement close to DSM resolution

kikislater:
Need to be tested before merging but It should work !


pierotofy:
Thanks @kikislater! Have you tested it also? You can do so by launching WebODM via `./webodm.sh start --dev` from your fork.

pierotofy:
Works, thanks @kikislater !

kikislater:
Ok thank you, I didn't get time to test yet but I tested in grass with python script !
Thank you. Huge improvement IMO

PR #851, opened 2020-04-13
Update README.md

pierotofy:
Added to the troubleshooting section - "On Windows, the storage space shown on the WebODM diagnostic page is not the same as what is actually set in Docker's settings. | From Hyper-V Manager, right-click “DockerDesktopVM”, go to Edit Disk, then choose to expand the disk and match the maximum size to the settings specified in the docker settings. Upon making the changes, restart docker."


pierotofy:
Thanks @Vantage-DS ! :+1:

PR #856, opened 2020-04-25
Change detection plugin

nchamo:
Hey ODM maintainers!

I have a new plugin that you might like to include into WebODM, it's about change detection. The plugin compares two tasks on the same project, by substracting the different tasks and displaying the differences.

Since it's impossible to make something that would work for everyone, the plugin is highly configurable.

Screenshot1

You can set the target resolution, the minimum difference height and area for differences to be considered changes.

Also, there are two preview modes, contours and heatmap. The first one displays changes as contours, and there is some extra information displayed when you click each one, such as min, max, avg and std diff height.

Screenshot2

The second one, just displays the 'raw' difference values as a heatmap. 

Screenshot3

Finally, there is something that you might notice when taking a look at the code. I left some code that is not currently being used that would allow users to align the two tasks before detecting changes. We discussed a little bit about in on the [forum](https://community.opendronemap.org/t/how-to-implement-change-detection/3821/4), but basically the idea is to first align the two clouds using ICP, and then regenate the DEMs for the new aligned compare cloud. I haven't gotten the time to work on it yet.

Until that code is in place, all tasks used on the plugin will need to have both a DSM and DTM. The align feature is currently disabled on `main.js`. It is called `alignSupported`.

So let me know what you think! Let me know if you need datasets to compare :)


pierotofy:
This is awesome! 

Thanks @nchamo !

plugins/changedetection/compare.py
```diff
@@ -0,0 +1,139 @@
...
+    reference_dtm_array = reference_dtm.read(1, masked=True)
+    compare_dsm_array = compare_dsm.read(1, masked=True)
+    compare_dtm_array = compare_dtm.read(1, masked=True)
+
+    # Calculate CHMs
+    chm_reference = reference_dsm_array - reference_dtm_array
+    chm_compare = compare_dsm_array - compare_dtm_array
+
+    # Calculate diff between CHMs
+    diff = chm_reference - chm_compare
```
limwenqing, line 32: should be 
diff = abs(chm_reference - chm_compare)
right? with abs, whos reference and whos compare will not be a matter, it will detect any "changes"
unlike now, either higher or either lower, it is weird.

limwenqing:
Hi @nchamo, it is really a great plugin!
I have tested the change detection plugin. However, the results are not that satisfying. Maybe it is my dataset problem. Anyway, this method highly relies on the quality of the DTM. Will taking orthophoto into account be better? May I know hows your progress on the ICP?

plugins/changedetection/compare.py
nchamo: On the first version I actually did that, but it didn't work as expected. I realized it was much more useful to just be able to determine the reference and compare tasks, instead of simply calculating the absolute difference. That way, you can determine what data is more important, and it is easier to ignore noise.

nchamo:
Hey @limwenqing !

Indeed, this method relies heavily on having a good DTM. If that is not the case for you, then it might help to use the [ground rectification tool](https://github.com/OpenDroneMap/ODM/pull/1083). It certainly helped in my case.

Regarding ICP, I did some testing on my fork, and the improvement wasn't significant. If both tasks were made by the same drone, then using ICP will just make the detected changes more defined on the edges, but nothing more. So I didn't try to add it to the ODM project, because it would take a lot of work to do it properly, instead of the hacky way I did it.

If you want to implement ICP by yourself, it should be really easy to integrate it with the existent plugin. [Here ](https://community.opendronemap.org/t/how-to-implement-change-detection/3821)is a discussion thread on the forum, that might help you.

limwenqing:
Thanks for your explanation @nchamo .

Didn't notice that got such option, I will try out the ground rectification.
Thank you!

usman-rind:
Hello Guys, 


https://github.com/OpenDroneMap/WebODM/assets/60088877/18e3ba79-e7ca-4796-8696-fd95bbfc6d09

We are using DJI Mavic 3 M and stitching multispectral images but the bands of the stitched image are not aligned as you can see in the attached video. 

could you please help why this is happening and how can we fix that? 
Is any of the above mentioned plugin or technique that can be used or any parameters of ODM that can be used to improve and reduce or eliminate this misalignment of different bands.

usman-rind:
> Hello Guys,
> 
>  WhatsApp.Video.2024-01-12.at.21.03.26.mp4 
> We are using DJI Mavic 3 M and stitching multispectral images but the bands of the stitched image are not aligned as you can see in the attached video.
> 
> could you please help why this is happening and how can we fix that? Is any of the above mentioned plugin or technique that can be used or any parameters of ODM that can be used to improve and reduce or eliminate this misalignment of different bands.

Guys, we further tested stitching on the Pix 4d Fields and there is no misalignment within the bands, could you please how this can be fixed in ODM. is there any related open issue?

smathermather:
@usman-rind -- can we start this conversation at community.opendronemap.org. Of if you have an example dataset with what you see as a bug that needs squashed or enhancement needed for processing, please open a separate issue for that. Thanks!

PR #861, opened 2020-05-08
Super minor fixes to plugins

nchamo:
We are making super minor fixes to two different plugins.

First, the change detection plugin. On that case, the stored value is a string. So, we are converting it into a boolean when reading it.

Then, on the elevation map plugin, apparently `round` can return floats in same cases. It was reported on https://community.opendronemap.org/t/problems-with-elevation-map-and-contours-plugins-on-latest-native-build-on-16-04/4171


pierotofy:
Looks good (as usual) :+1: thanks @nchamo !

PR #888, opened 2020-07-07
Update `_task.md` with missing endpoint: `POST /tasks/import`

RGrze:
This pull request introduces update for API documentation.

There was an undocumented endpoint in `Task`, more precisely `POST /api/projects/{project_id}/tasks/import`, so description of the endpoint was added.


pierotofy:
Looks great! Thanks @RGrze :+1:

PR #889, opened 2020-07-15
adjusting python DB settings to use environment variables

IslamHeggy:
We are trying to create a WebODM kubernetes deployment.

So in our case we are creating a template, this template generates a pod that contains all the containers running using your mentioned docker-compose files.

The problem is that we need to address the database host in our case with localhost instead of the hard coded db value and maybe external DB in the future

We created a new image and added a hard coded localhost in the python settings to override this problem.

So in order not to keep a separate version of your image, This pull request contains replacing Database env variables while preserving the default values as they are.

### DIFF
--- a/webodm/settings.py
+++ b/webodm/settings.py
@@ -145,12 +145,12 @@ WSGI_APPLICATION = 'webodm.wsgi.application'
 
 DATABASES = {
     'default': {
-        'ENGINE': 'django.contrib.gis.db.backends.postgis',
-        'NAME': 'webodm_dev',
-        'USER': 'postgres',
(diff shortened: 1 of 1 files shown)


pierotofy:
Looks great, thanks @IslamHeggy :+1:

pierotofy:
There just seems to be a comma missing from line 48. After that's resolved we can merge this.

IslamHeggy:
Added the comma missing on the first line

IslamHeggy:
removed extra comma in the port line

pierotofy:
Thanks @IslamHeggy :pray:

PR #892, opened 2020-07-25
Adding GitHub integration to Cloud Import plugin

nchamo:
Hey everyone!

On this PR, I'm adding GitHub integration for the `Cloud Import` plugin!

Had to make a few minor changes to the platforms framework, but it ended up being pretty straight forward. I tested it on some of the ODM datasets.

Also, you may notice some changes on the `webodm/settings.py` file. I don't think these particular changes should be merged, but I'm not sure how fix the problem I'm having. I already explained it on the [forum](https://community.opendronemap.org/t/problem-with-celery-task/4980).

Let me know what you think 😄

### DIFF
--- a/app/plugins/worker.py
+++ b/app/plugins/worker.py
@@ -1,5 +1,29 @@
+import inspect
 from worker.celery import app
 # noinspection PyUnresolvedReferences
 from worker.tasks import execute_grass_script
 
 task = app.task
+
+def run_function_async(func, *args, **kwargs):
+    """
+    Run a function asynchronously using Celery.
+    Plugins should use this function so that they don't
+    have to register new Celery tasks at startup. Functions
+    should import any required library at the top of the function body.
+    :param {Function} a function to execute
+    """
+    source = inspect.getsource(func)
+    return eval_async.delay(source, func.__name__, *args, **kwargs)
+
+
+@app.task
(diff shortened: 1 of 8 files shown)


pierotofy:
Hey @nchamo :hand: this is awesome. I'm going to look over the Celery problem and see what kind of stuff we can add in WebODM to make this more generic.

pierotofy:
Hey @nchamo :wave: I've added a new `run_function_async` which is able to run a generic function on a Celery worker. This way the plugin can just invoke `run_function_async` and doesn't have to worry about registering with Celery at boot time (simplifies things quite a bit in the scope of plugin development).

Let me know if this works for you.

Feel free to merge it if the plugin works correctly after these changes (I did some basic tests and seems to work, but double check). :+1:

nchamo:
Hey @pierotofy !

Thanks for the help! Just tried it and it seems to be working, so I'll go ahead and merge.

Just as an FYI, there are two other plugins that may need the same fix:
* [Open Aerial Map](https://github.com/OpenDroneMap/WebODM/blob/master/plugins/openaerialmap/api.py#L125)
* [Cesium Ion](https://github.com/OpenDroneMap/WebODM/blob/master/plugins/cesium-ion/api_views.py#L360)

Issue #894, opened 2020-08-03
Rasterio 1.1.5 / GDAL 3.1 displays black background in tiler for multispectral datasets (bands > 3)


pierotofy:
This did not happen with rasterio 1.1.10 / GDAL 2.x

First reported on https://community.opendronemap.org/t/wierd-multispectral-image-plotting-in-webodm/5044/4

PR #895, opened 2020-08-03
Workaround for black background in multispectral datasets in tiler #894

pierotofy:
Workaround for #894 

This is less than ideal, but should work for most use cases. Setting a nodata value could potentially hide valid raster cells with a value of zero.

### DIFF
--- a/app/api/tiler.py
+++ b/app/api/tiler.py
@@ -168,7 +168,12 @@ class Metadata(TaskNestedView):
                 if has_alpha_band(src):
                     band_count -= 1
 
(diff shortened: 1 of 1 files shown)


pierotofy:
image

smathermather:
Is the underlying issue the inability to set a transparency band due to having so many channels?

pierotofy:
I have no idea of the underlying cause. :man_shrugging: 

Maybe a bug in rasterio. Maybe in GDAL's VRT driver.

smathermather:
GDAL's VRT driver is buggy I have noticed in other projects.

PR #897, opened 2020-08-03
Change Detection: fix export by adding appropriate import

nchamo:
Hey everyone!

Just detected that only the GeoJSON export worked, because I was missing an `import`.

That's it, super small haha


pierotofy:
Nice catch! Thanks.

PR #922, opened 2020-10-12
Auth0

SamuelBradley:
Fixed a bug with the logout behaviour and made sure mapping of Auth0 'name' field conformed to Django's 30 character limit for 'first_name'


SamuelBradley:
hmm this shouldn't of ended up here, will try and delete

PR #926, opened 2020-10-21
Fix #866 - Hash password while creating user through admin API

### LINKED ISSUE #866: Login fails for user created via API (password not processed or lacking documentation)

Chrome 81 on macOS (Catalina)

### What is the problem?

When creating a user via the API, the password does not work. It seems the API is not correctly processing the password, or the documentation doesn't state in which format the password should be provided.

I expect the login credentials to work.

First generate an API auth token using your admin credentials:

```
curl -X POST -d "username=adminuser&password=adminpass" http://localhost:8000/api/token-auth/
```

Then invoke the "create a user" API endpoint:

```
curl -X POST -H "Authorization: JWT <token>" -d "username=testuser&password=testpass" http://localhost:8000/api/admin/users/
```

Here is the response. Note that the password is returned in cleartext:

```
{"id":10,"password":"testpass","last_login":null,"is_superuser":false,"username":"testuser","first_name":"","last_name":"","email":"","is_staff":false,"is_active":false,"date_joined":"2020-05-22T09:40:14.285065Z","groups":[],"user_permissions":[]}
```

When checking WebODM it shows that the password is in an invalid format:

image

I'm guessing the API is not processing the password (i.e. it is assuming it is in whatever internal format webodm uses to stored the salted/hashed password).

Is this a bug or a feature?

The API docs don't state anything about the format. Just says "password".

https://docs.webodm.org/#create-a-user

abrbhat:


abrbhat:
Seems travis errored in set up

pierotofy:
Hey @abrbhat  :hand: this is great!

Try merging the changes from #927 and see if the build gets fixed :+1:

abrbhat:
Seems good now

pierotofy:
Thanks for the fix @abrbhat :+1:

abrbhat:
Glad to help :)

PR #941, opened 2020-12-14
Localization and Translation

pierotofy:
Resuming work on the `i10n` branch that was started years ago.

Goal is to get localization and translation support in WebODM. Lots of stuff to do.


smathermather:
Woah! Awesome.

pierotofy:
Looks promising: 

image

pierotofy:
Going to re-open a different PR so we can get the weblate efforts going right away.

smathermather:
I might just default to _Italian_ because.

pierotofy:
I was thinking of adding this to my default: https://www.technomancy.org/python/pirate-po-l10n-i18n-translate-your-site/

PR #968, opened 2021-03-19
Cesium-ion plugin fix

mateo3d:
The cesium-ion plugin needed to be renamed cesiumion to allow module imports using the new celery task handler. Moved upload-only functions to uploader.py for use with the getSource function. 

- 3d models do not import with correct location yet - they are pretty close but not perfect.
- Tile button does not appear with Cloud Import enabled (only on projects that actually use cloud import)


pierotofy (approved):

pierotofy:
Thanks @mateo3d ! :clinking_glasses: 

The 3D model alignment *might* be due to the changes we've made in https://github.com/OpenDroneMap/ODM/pull/1233, the textured model is aligned to UTM coordinates minus the offset specified in `odm_georeferencing/coords.txt`. Maybe this helps a future PR.

Issue #891, opened 2020-07-22
Rewrite assure_cogeo to use GDAL 3.1 (needs GDAL update)


pierotofy:
assure_cogeo will hang indefinitely while processing certain .TIF files, on certain machines. Related: https://community.opendronemap.org/t/processing-hangs-at-compressing-all-zip/2013/25

Attempting to figure out a fix for rio-tiler/rasterio led to a dead end (see forum thread). GDAL >= 3.1 has native support for writing Cloud Optimized GeoTIFFs.

PR #972, opened 2021-04-06
Use ubuntu 20.04 as base to ensure gdal version is >= 3.1

### LINKED ISSUE #971: WebOdm hangs on compressing all.zip for some datasets/runs.
Chrome 89 on Linux

### What is the problem?

WebOdm hangs on compressing all.zip for some datasets/runs.
This issue seems very similar to what is reported here: https://community.opendronemap.org/t/processing-hangs-at-compressing-all-zip/2013/28
which should have been fixed by this pull request: https://github.com/OpenDroneMap/WebODM/pull/893
which references that gdal should be updated to 3.1. However on the newest master it seems that GDAL is no longer 3.1 but 3.0.
Running "dpkg -l | grep gdal" in the worker container gives "gdal-bin 3.0.4+dfsg-1~bionic0"
I also see a lot of these warnings:
WARNING Using legacy implementation (GDAL >= 3.1 not found)
when running "docker logs worker"

I am not that familiar with webodm, but as far as i can see this could be caused by this commit https://github.com/OpenDroneMap/WebODM/commit/ff989cc3203a79dfa62146789ed2842715cbd6ed#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557
which changes the base from debian to ubuntu. (In the original issue it was mentioned that gdal 3.1 was depending on debian).

WebOdm succesfully completing.

I cannot reproduce this every time, it only happens on some particular runs of the same data set. retrying from scratch a couple of times eventually makes it finish.

cwitting:
In ubuntu 18.04 the gdal version in ppa:ubuntugis/ubuntugis-unstable is only 3.0, but in 20.04 it is 3.2.1.

This should fix #971 and is related to #891 which needs gdal >= 3.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 MAINTAINER Piero Toffanin <pt@masseranolabs.com>
 
 ENV PYTHONUNBUFFERED 1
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @cwitting ! Looks great.

PR #1007, opened 2021-06-24
Improved performance of camera shot markers 

zaakzaku:
I've processed around 1000 images. I wanted to show camera shot images but the browser struggle in rendering thousands of markers. 
struggle

### DIFF
--- a/app/static/app/js/components/Map.jsx
+++ b/app/static/app/js/components/Map.jsx
@@ -221,11 +221,6 @@ class Map extends React.Component {
 
             // Add camera shots layer if available
             if (meta.task && meta.task.camera_shots && !this.addedCameraShots){
-                const cameraMarker = L.AwesomeMarkers.icon({
-                    icon: 'camera',
-                    markerColor: 'blue',
-                    prefix: 'fa'
-                });
 
                 const shotsLayer = new L.GeoJSON.AJAX(meta.task.camera_shots, {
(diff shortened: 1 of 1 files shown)


pierotofy:
Hey @zaakzaku :hand: thanks for the PR! Looks great. :pray:

Saijin-Naib:
This is really neat! Can you describe what you changed and why it helps so I can understand better?

Is it not drawing all the markers all the time, but instead culling them so it only draws the visible ones?

zaakzaku:
> This is really neat! Can you describe what you changed and why it helps so I can understand better?
> 
> Is it not drawing all the markers all the time, but instead culling them so it only draws the visible ones?

Thank you for your question,

The performance issue is due to the fact that each marker is an individual DOM element. Browsers struggle in rendering thousands of them.
In this case, an easy workaround would be instead to use Circle Markers and have them rendered on a Canvas.
`
 new L.CircleMarker(latlng, {
       color: '#3498db',
       fillColor: '#3498db',
       fillOpacity: 0.9,
       radius: 10,
       weight: 1
       });
`
With this solution, each Circle Marker is no longer an individual DOM element but instead is drawn by Leaflet onto a single Canvas, which is much easier to handle for the browser.

Saijin-Naib:
That's really neat, and an elegant solution!

PR #1009, opened 2021-07-01
Short links plugin

pierotofy:
Adds a plugin to create short links when sharing a map/3d model display link. (Makes it easier to type)﻿


pierotofy:
image

PR #1016, opened 2021-07-08
Update formulas.py

Chihiro-KATO:


pierotofy:
Hey @Chihiro-KATO :hand: thanks for the PR! Happy to add a new filter, but can you leave the one you removed? Other people might need it.

Chihiro-KATO:
I wasn't sure if increasing the number of "camera-filters" items would cause any problems in the program's operation, so, for the purpose of checking the program, I simply overwrote the filter items without increasing the number of items.

If there is no problem with increasing the number of items, I will create another pull request.

pierotofy:
Looks great, thanks! :pray:

Issue #1015, opened 2021-07-07
WebODM on macOS stop working.


PLSomebody:
macOS 11.4 (20F71), Safari 14.1.1 (16611.2.7.1.4)

Suddenly WebODM stops working.

When I'm using ./webodm.sh start 
I'm getting this:
Checking for docker...   OK
Checking for git...   OK
Checking for docker-compose...   OK
Starting WebODM...

Using the following environment:
================================
Host: localhost
Port: 8000
Media directory: appmedia
SSL: NO
SSL key: 
SSL certificate: 
SSL insecure port redirect: 80
Celery Broker: redis://broker
Default Nodes: 1
================================
Make sure to issue a ./webodm.sh down if you decide to change the environment.

docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml start || docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up --scale node-odm=1

And nothing more. Trying to open localhost:8000 effects in message that browser can't connect to the host.
I was trying to reinstall Docker, WebODM - no change in behavior.

pierotofy:
Hi @PLSomebody,

It looks like there is some information missing from your issue that will be needed in order to diagnose and fix the problem at hand. Please take a look at the [Issue Template](https://github.com/OpenDroneMap/WebODM/blob/master/ISSUE_TEMPLATE.md), which will tell you **exactly** what your issue has to contain in order to be processable.

Also, double check that this is the right place. If you are just asking for information, reporting feedback or proposing a few feature, the right place to ask is the [Community Forum](http://community.opendronemap.org/c/webodm), not here.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next week (until 2021-07-08 19:05) I'll close this issue so it doesn't clutter the bug tracker.

Cheers!
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

sushidub:
I'll see if I can provide what's needed here as I'm having almost the exact same issue as the original issuer.

### What's your browser and operating system?
Chrome 91 on macOS (Catalina)

### What is the problem?
As of a few weeks ago all of the following `start` variations stopped instantiating any builds of the respective docker images needed to run WebODM/NodeODM and subsequently won't run any containers -- all while failing ever so silently.

[...]

PR #1024, opened 2021-07-18
per #1015 discussion - this removes redundant docker 'start' command

sushidub:
Removes redundant docker 'start' command from start function as docker 'up' handles the former.

Issue #1015 

_NOTE: Might be irrelevant considering the type of change but I was unable to test as `--dev` therefore unable to run tests due to the webapp container perpetually restarting when executing start or restart with --dev flag (culprit seemed to be `ModuleNotFoundError: No module named 'rio_tiler'`. I can provide that full specific log if you want._

### DIFF
--- a/webodm.sh
+++ b/webodm.sh
@@ -265,7 +265,7 @@ start(){
 		echo "Will enable SSL ($method)"
 	fi
 
-	command="$command start || $command up"
+	command="$command up"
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Thanks @sushidub ! :+1:

PR #1027, opened 2021-07-27
Fix camera animation scene state trap

### LINKED ISSUE #1025: Bug: Unable to remove PoTree Camera Animations
STR:
1. Create project with pointcloud
2. Add default camera animation
3. Add more
4. Select animation and delete key frames
5. See that empty animations can not be purged
6. (Possible breakage of PoTree view from above steps that persists across app instances)

pierotofy:
Closes #1025﻿

### DIFF
--- a/app/static/app/js/ModelView.jsx
+++ b/app/static/app/js/ModelView.jsx
@@ -279,6 +279,7 @@ class ModelView extends React.Component {
       // scene updates when these change.
       delete json.view;
       delete json.settings;
+      delete json.cameraAnimations;
 
       return json;
   }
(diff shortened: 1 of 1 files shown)


Saijin-Naib:
Possible unfortunate side-effect of not being able to save _**desired**_ camera animations once you leave the 3D view, which might be a papercut if someone painstakingly creates an animation only to find it gone the next time they return to the project.

Saijin-Naib:
Follow up:

Default camera position seems to be unset, instead of defaulting to Full Extent.

PR #1028, opened 2021-07-28
Hide set camera view in public URLs

pierotofy:
- Hides the set camera view button from shared URLs.﻿


pierotofy:
* Added editable short links plugin

pierotofy:
image

PR #1030, opened 2021-07-29
Add additional statistics on task items in dashboard

pierotofy:
Adds additional useful information from the report's statistics directly on the dashboard.


pierotofy:
image

Issue #1039, opened 2021-08-25
Point cloud not appearing for some models in potree


OKaluza:
### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Firefox 91 on Ubuntu Linux

### What is the problem?

No point cloud displayed in the 3d model view, the textured model loads fine if selected.

### What should be the expected behavior?

Point cloud should be viewable

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? Please include multiple screenshots of the problem! If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

It seems to occur with small reconstructions over small areas, eg: less than 10 metres.

I have fixed the bug with a simple modification to potree so will make a pull request

PR #1040, opened 2021-08-25
Fix potree bug when info.srs is undefined

OKaluza:
Causing point cloud to not be displayed for certain data sets

I have created a pull request in potree for this fix, but it might take a while to get merged so added it here directly too

Solves this issue: #1039


pierotofy:
Thanks @OKaluza ! We'll merge it here in the meanwhile :)

PR #1043, opened 2021-08-27
updated docker-compose files to v3

mostafaradwan:
- Updated docker-compose files to v3
- Added 3 replicas to the WebODM-webapp based on feedback from the American Redcross Team(ARC)
- This is part of a JIRA request from the code4good.io project at the ARC

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -1,6 +1,6 @@
 # This configuration builds from Dockerfiles 
 # instead of pulling from opendronemap hub
-version: '2.1'
+version: '3.9'
 services:
   db:
     build: ./db
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,4 +1,4 @@
-version: '2.1'
+version: '3.9'
 services:
   webapp:
     entrypoint: /bin/bash -c "chmod +x /webodm/*.sh && /bin/bash -c \"/webodm/wait-for-postgres.sh db /webodm/wait-for-it.sh -t 0 broker:6379 -- /webodm/start.sh --create-default-pnode --setup-devenv\""
--- a/docker-compose.nodemicmac.yml
+++ b/docker-compose.nodemicmac.yml
@@ -2,7 +2,7 @@
 # a default processing node instance. This is best for users
 # who are just getting started with WebODM.
 
-version: '2.1'
+version: '3.9'
 services:
   webapp:
     depends_on:
--- a/docker-compose.nodeodm.yml
+++ b/docker-compose.nodeodm.yml
@@ -2,7 +2,7 @@
 # a default processing node instance. This is best for users
 # who are just getting started with WebODM.
 
-version: '2.1'
+version: '3.9'
 services:
   webapp:
     depends_on:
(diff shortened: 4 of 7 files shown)


mostafaradwan:
Thanks for your attention @pierotofy.  That's a great point. 

Would it make sense to add it under a separate branch entitled (docker-compose-v3) and add a note in the README file?

danbjoseph:
What if the v3 compose file is added but with a different file extension that keeps it from getting read by the docker commands, and the README notes that you'll need to switch the filenames to use the v3 file? 
Or would it be possible to adjust [webodm.sh](https://github.com/OpenDroneMap/WebODM/blob/master/webodm.sh) to use `docker-compose.yml` (v2) by default but accept a paramater that triggers the use of a `docker-compose.v3.yml`?

Regarding the "3 replicas" - our concern was just that the WebODM instance didn't seem to automatically restart after a system restart, and we were looking for either a code change or documentation to help us have WebODM come back up after the server itself was rebooted.

pierotofy:
A separate branch would work also.

danbjoseph:
A separate branch would require regular minor PRs to pull updates from the overall project over into the branch?
Does `replicas` even solve the issue of WebODM restarting after a system restart? From the [docs](https://docs.docker.com/compose/compose-file/compose-file-v3/#replicas):
> If the service is replicated (which is the default), specify the number of containers that should be running at any given time.

How does running more containers help restart WebODM after a system restart?

pierotofy:
> A separate branch would require regular minor PRs to pull updates from the overall project over into the branch?

Yes. So maybe not ideal. 

It would be interesting to hear more about the core issue that was being fixed/solved here. Maybe there are alternatives that don't require a migration to v3.

danbjoseph:
It looks like the only big change with the upgraded docker-compose file is adding the following:
```
    deploy:
      mode: replicated
      replicas: 3
```
My concern was just that the WebODM instance didn't seem to automatically restart after a system restart, and I was looking for either a code change or documentation details to help us have WebODM come back up after the server itself was rebooted. It is unclear to me if `replicas` will solve the issue of WebODM restarting after a system restart. From the [docs](https://docs.docker.com/compose/compose-file/compose-file-v3/#replicas): "If the service is replicated (which is the default), specify the number of containers that should be running at any given time." It is not clear to me how running more containers will help restart WebODM after a system restart.

mostafaradwan:
@danbjoseph  Thanks for looking into this.

It's not clear to me what system restarts are. Is that the entire ODM stack? 

By having more replicas WebODM (the front-end component) should be more resilient and highly-available by having more than one container to handle requests in case one dies or restarts at any point.

The way to find out is to run a quick test, trigger a system restart, and observe if having 3 replicas resolves the issue.

Is that something you can try and report the results here?  Let me know if you've any questions.

danbjoseph:
For example if using AWS and the EC2 instance is rebooted. Or local hardware running it such as an Intel NUC is turned off and back on.

mostafaradwan:
@danbjoseph: I was under the impression this has to do with WebOEM container crashing sometimes and that's the point of adding the replicas. 

If you encounter rebooting issues of the underlying VMs (i.e: EC2 instances), the docker engine should restart as well and take care of recreating the containers. If Docker is not configured to start on boot, this is very easy to configure per the steps in the links below.

**Configure Docker to start on boot**
https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot

Please let me know your findings.

pierotofy:
I'm going to close this PR for the time being, as this might be better solved by making infrastructure configuration changes rather than software changes. :pray:

PR #1054, opened 2021-09-07
add celery result backend config: retry policy

yudeqang:
Sometimes  celery redis connection will be abnormal，this is error log:
image
Add this options will fix it.


pierotofy (approved):
Thanks @yudeqang  ! :pray:

PR #1066, opened 2021-10-05
Debian package to install/configure WebODM, including NVIDIA and Intel GPU support.

bluecamel:
This PR adds scripts to build a Debian package, which installs and configures WebODM.  It attempts to detect available GPU hardware and customizes the installation accordingly.  GPU detection supports NVIDIA and Intel, but doesn't currently support AMD, as I don't have any hardware to test.

There is a [companion PR that modifies the nodeodm docker image to support Intel’s OpenCL drivers](https://github.com/OpenDroneMap/NodeODM/pull/170).


dpkg/build.sh
```diff
@@ -0,0 +1,73 @@
...
+
+function prepareDebDir() {
+    if [ -d "${DEB_DIR}" ]; then
+        rm -Rf "${DEB_DIR}"
+    fi
+
+    mkdir -p "${DEB_DIR}"
+}
+
+function prepareNodeODMDir() {
```
bluecamel, line 58: This is temporary, and can be removed after the [NodeODM PR](https://github.com/OpenDroneMap/NodeODM/pull/170) merges.

dpkg/webodm/DEBIAN/postinst
```diff
@@ -0,0 +1,131 @@
...
+        if [ -z "${SED_OUT}" ]; then
+            sed -i "/^\s\{4\}image/a \ \ \ \ group_add:\n\ \ \ \ \ \ - ${gid}" "${DOCKER_COMPOSE_FILE}"
+        fi
+    done
+}
+
+function buildNodeODMGPUIntelDockerImage() {
+    cd "${NODEODM_DIR}"
+
+    # We won't have to do this after the image is on Docker Hub.
```
bluecamel, line 31: This is temporary, and can be removed after the [NodeODM PR](https://github.com/OpenDroneMap/NodeODM/pull/170) merges.

detect_gpus.sh
```diff
@@ -0,0 +1,30 @@
...
+        return
+    fi
+
+    lspci | grep "VGA.*Intel"
+    if [ "${?}" -eq 0 ]; then
+        export GPU_INTEL=true
+        return
+    fi
+
+    # Total guess.  Need to look into AMD.
```
bluecamel, line 19: Maybe I should remove this until someone can add AMD?

service/webodm-docker.service
```diff
@@ -8,6 +8,7 @@ Type=forking
 User=odm
 Group=odm
 WorkingDirectory=/opt/WebODM
+Environment=PATH=/home/odm/.local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
bluecamel, line 11: This insures that the service uses the version of docker-compose installed by pip, which is much newer than the version available from Ubuntu repositories.

pierotofy:
Thanks @bluecamel ! This looks like a great effort.

The only thing to adjust before I merge this is graceful handling on non-Linux systems. Neither Windows (via mingw64) or macOS have the `lspci` command, so we're breaking ./webodm.sh for these two platforms. Perhaps some conditional checks in `detect_gpus.sh`, or just moving the `detectGPUs` function to `webodm.sh` and then checking for $platform == "Linux" (https://github.com/OpenDroneMap/WebODM/blob/master/webodm.sh#L6) would do the trick.

bluecamel:
> Thanks @bluecamel ! This looks like a great effort.
> 
> The only thing to adjust before I merge this is graceful handling on non-Linux systems. Neither Windows (via mingw64) or macOS have the `lspci` command, so we're breaking ./webodm.sh for these two platforms. Perhaps some conditional checks in `detect_gpus.sh`, or just moving the `detectGPUs` function to `webodm.sh` and then checking for $platform == "Linux" (https://github.com/OpenDroneMap/WebODM/blob/master/webodm.sh#L6) would do the trick. We can just disable GPU stuff on non-Linux platforms for the moment.

Ah, I missed that!  I'll update and test on Windows.  👍🏼

bluecamel:
@pierotofy Sorry, that was a very naive promise for me to make.  I verified this change on Linux, but will take me a while to get another OS ready to test.

BTW, I kept the script separate because it's also used by the install script.

bluecamel:
> Because I didn't build the nodeodm:gpu.intel image locally. We don't expect our users to do so, maybe a GitHub action that automatically builds the GPU image for NodeODM needs to be added?

Yeah, right now, the install script builds that.  I [added a workflow](https://github.com/OpenDroneMap/NodeODM/pull/171) that hopefully should do it.  After that's on DockerHub, I can update this PR to remove that.

bluecamel:
I may need to find another way to do that. 🤔

bluecamel:
Well, I tried get rid of an extra build entirely, but couldn't find a good solution, so ended up just breaking out the host-specific group to another image that can be built from `webodm.sh`.  Since the main image will already be pulled from DockerHub, this build should be really quick.  Definitely open to other ideas.

bluecamel:
> One idea that doesn't require a build could be to add a `run_gpu.sh` script in NodeODM (maybe under `/var/www/run_gpu.sh`) that checks whether the proper RENDER_GROUP_ID has been set. You can pass environment variables to the container at runtime via docker-compose (we do something like that for webodm.sh) with the proper value. Then you can execute the `groupadd`, `usermod` commands at runtime before launching the node process `/usr/bin/node /var/www/index.js` (just remember to propagate the command line arguments passed from the docker command through `run_gpu.sh` to `node index.js`).
> 
> You might have to keep the container running as root until run_gpu.sh is executed (change `USER odm` to `USER root`), since `odm` probably doesn't have permission, but you can then run the node process using the odm user by giving up root.

Okay, I have this working now.  It was a tricky to get the environment variables passed through, but I think it's good.

PR #1067, opened 2021-10-06
Rio update

markFieldman:
Reopen as merge to master. Thanks for the support!


pierotofy:
Thanks for the PR! 

There's one more problem I've found: if you call the tile API without a `?rescale` parameter you get a crash:

```
http://localhost:8000/api/projects/1/tasks/80f0b969-ee13-412d-9c5d-5c917b4bcee3/orthophoto/tiles/21/512673/738967.png
```

You can reproduce this by taking the URL of a tile request and removing the `?rescale` parameter.

markFieldman:
> Thanks for the PR!
> 
> There's one more problem I've found: if you call the tile API without a `?rescale` parameter you get a crash:
> 
> ```
> http://localhost:8000/api/projects/1/tasks/80f0b969-ee13-412d-9c5d-5c917b4bcee3/orthophoto/tiles/21/512673/738967.png
> ```
> 
> You can reproduce this by taking the URL of a tile request and removing the `?rescale` parameter.
> 
> ```
> AttributeError
> 'NoneType' object has no attribute 'split'
> /webodm/app/api/tiler.py in get, line 370
> ```
> 
> For orthophotos, there should be no rescaling if the rescale parameter is not passed.

Can we use default rescale like (0,256) in this case?

pierotofy:
Good point, yes I set it to 0,255 https://github.com/OpenDroneMap/WebODM/pull/1067/commits/3d8d5a705f51ca1cfb158db255c9d742900b79bb by default

pierotofy:
Looks good! I've made some small changes to avoid introducing matplotlib, but otherwise it's looking pretty solid.

Thanks so much for your contribution @markFieldman :pray: it looked like a fairly large effort.

vincentsarago:
❤️ also FYI I'm actively working on rio-tiler V3 (after releasing new morecantile and rio-cogeo). Feel free to 👀 and 👋

markFieldman:
> ❤️ also FYI I'm actively working on rio-tiler V3 (after releasing new morecantile and rio-cogeo). Feel free to 👀 and 👋

Thanks for your awesome product! Also thanks for you awesome documentation - it was so helpful! I filled very comfortable when works with rio-tiler 🙏

PR #1068, opened 2021-10-07
Added support of getting tiles and metadata with geoJSON polygon specified at request param

holyCowMp3:
Added support of retrieve metadata for cutline specified as GeoJSON feature or polygon in boundaries ~header~
UPD - migrate this functionality to the request param "boundaries", instead of Header. Also added support of cutting geoJSON tiles using same functionality

### DIFF
--- a/app/api/tiler.py
+++ b/app/api/tiler.py
@@ -1,5 +1,4 @@
 import json
-import base64
 import numpy
 import rio_tiler.utils
 from rasterio.enums import ColorInterp


markFieldman:
I think, we can adopt this feature to the standalone boundaries geoJSON, which must be saved at the server-side alongside processed geoTIFF`s

markFieldman:
> Yep, perhaps by adding a new field in the `app/models/task.py` model. Is this ready for review/merge or do you need to do more additions?

Yes, It is ready for review.

PR #1070, opened 2021-10-11
Added possibility to import file from system.

holyCowMp3:
Have an issue during trying to import files from the file system. Instead of using requests file adapter - use native copyfile to unpack filesystem zip with assets.
Format of url for usage is - `file:///this/is/path/to/all.zip`


pierotofy:
I'm a bit on the fence about allowing arbitrary system URLs to be imported; this is probably a security vulnerability. We can allow uploads from the filesystem, but should restrict it to a single subfolder folder (e.g. `<media-dir>/imports` ?)

markFieldman:
Ok, i will add security checks for this scenario

PR #1109, opened 2021-12-08
Update boot.py - Add auto-boundary

Saijin-Naib:
I think it might be cheap insurance to have it on for all presets, as it certainly can help some datasets that would otherwise fail to reconstruct to reconstruct properly. 

It has little benefit for at-nadir surveys and processing templates (and I've observed no cost/loss in testing), but sometimes these datasets can have unexpectedly large reconstruction areas above/below the plane of reconstruction due to things like water or other spectral material in the survey, and this should prevent that edge case from being fatal for processing.

Initially, I had it for "3D" focused presets, but added it for all due to the above reasoning. Welcome to reconsider my thought process here.

### DIFF
--- a/app/boot.py
+++ b/app/boot.py
@@ -100,48 +100,58 @@ def boot():
 def add_default_presets():
     try:
         Preset.objects.update_or_create(name='Multispectral', system=True,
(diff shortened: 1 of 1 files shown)


pierotofy:
I don't see harm in it and the reasoning makes sense.

Thanks! :pray:

PR #1118, opened 2021-12-22
Refined Boundary load MIME Types

Saijin-Naib:
Changed text/* to text/plain
Added application/geo+json to add GeoJSON support directly.


pierotofy (approved):
Nice addition, thanks! This will make it easier to select json files.

PR #1122, opened 2022-01-17
Add DroneDB integration (download & share)

HeDo88TH:
DroneDB settings page:

login

Dataset choice and upload

Schermata del 2022-01-17 02-08-59

Now working on refining the process and on DroneDB upload result files.


coreplugins/dronedb/api_views.py
```diff
@@ -273,7 +273,17 @@ def ddb_cleanup(sender, task_id, **kwargs):
     # When a task is re-processed, make sure we can re-share it if we shared a task previously
 
     logger.info("Cleaning up DroneDB datastore for task {}".format(str(task_id)))
-    #ds.del_key(get_key_for(task_id, "status"))
+
+    task = models.Task.objects.get(id=task_id)
```
HeDo88TH, line 277: This is not working. I need to get the project id from the task id. I'm using the ORM but it seems that no field `id` is found in the `project` object. What am I doing wrong? @pierotofy

HeDo88TH:
I'm now adding the logic to handle edge cases and cleanup. Both the download and the share workflow work! 🚀

HeDo88TH:
This is ready! @pierotofy

pierotofy:
* `celerybeat-schedule.db` should not be included in the repo.

pierotofy:
Using: `https://hub.dronedb.app/r/pierotofy/brighton-beach` as a URL improperly shows 

image

(20 instead of 18)

pierotofy:
I can choose a folder with no input images:

image

pierotofy:
Pressing "DroneDB" flashes a "Fill in your hub credentials" warning, even though my credentials are set (set a network throttle to see it better).

pierotofy:
Can we make the actual DroneDB icon instead of using the font-awesome cloud icon?

image

See the openaerialmap plugin for an example.

pierotofy:
`cameras.json` probably does not need to be uploaded automatically.

But other things like the PDF report or textured model are missing.

Should the user choose which items to upload?

pierotofy:
Really like the ability to browse organizations/datasets... it's a nice touch.

pierotofy:
Overall a few things need to be adjusted, but good progress!

pierotofy:
I'm not sure about the "Go To Dataset" button. I found it a bit confusing, perhaps we could remove it.

image

HeDo88TH:
Ready!

markFieldman:
Thanks for your efforts, @HeDo88TH!
So cool to see such wanted integration 😃

pierotofy:
[quote]
Thanks for this integration! Is it already usable?
[/quote]

Not yet, soon!

pierotofy (approved):

pierotofy:
Excellent, good to go!

PR #1125, opened 2022-01-24
Request a change about Nvidia detector

ichsan2895:
My previous pull is works perfectly in my computer, it can detect Nvidia driver. But I afraid my code don't work with another computer since I never test with that. So, I add my previous code (which worked in my computer) and keep the previous code (which I think it will works on another computer)


pierotofy:
Awesome, thanks!

PR #1132, opened 2022-02-01
webodm.sh: remove default_nodes and respect WO_DEFAULT_NODES from .env

### LINKED ISSUE #1131: webodm.sh should respect the WO_DEFAULT_NODES env variable by default
When starting WebODM with the command line (with `./webodm.sh start`) and after configuring the `.env` file, it appears that a node is started even if the variable `WO_DEFAULT_NODES` is set to 0.

Probably just adding `default_nodes=$WO_DEFAULT_NODES` after the `.env` sourcing [here](https://github.com/OpenDroneMap/WebODM/blob/6a16b1593c43d32d35edb802bed54e3aa1a0b4d7/webodm.sh#L39) would be enough.

gromain:
This fixes #1131 and harmonizes the way WO_DEFAULT_NODES is managed.

Also, I have a warning list from shellcheck for this script that I can fix while I'm at it. This will make the script more robust. For example it looks like the password does not support having spaces in it for now (see #1133 ).

```
Warning	Quote this to prevent word splitting. [SC2046]	3:16
Info	Not following: ./.env was not specified as input (see shellcheck -x). [SC1091]	32:8
Warning	Declare and assign separately to avoid masking return values. [SC2155]	59:12
Warning	Declare and assign separately to avoid masking return values. [SC2155]	68:12
Warning	Declare and assign separately to avoid masking return values. [SC2155]	73:12
Info	Check exit code directly with e.g. 'if mycmd;', not indirectly with $?. [SC2181]	173:8
Info	Check exit code directly with e.g. 'if mycmd;', not indirectly with $?. [SC2181]	180:8
Info	Check exit code directly with e.g. 'if mycmd;', not indirectly with $?. [SC2181]	187:8
Info	Check exit code directly with e.g. 'if mycmd;', not indirectly with $?. [SC2181]	195:8
Warning	Declare and assign separately to avoid masking return values. [SC2155]	207:11
Info	Double quote to prevent globbing and word splitting. [SC2086]	224:7
Info	Use -n instead of ! -z. [SC2236]	228:9
Info	Double quote to prevent globbing and word splitting. [SC2086]	251:7
Info	Double quote to prevent globbing and word splitting. [SC2086]	252:7
Error	> is for string comparisons. Use -gt instead. [SC2071]	280:29
Info	Use -n instead of ! -z. [SC2236]	299:8
Info	Use -n instead of ! -z. [SC2236]	303:8
Info	Use -n instead of ! -z. [SC2236]	311:8
Info	Use -n instead of ! -z. [SC2236]	311:34
Error	> is for string comparisons. Use -gt instead. [SC2071]	342:26
Info	Use -n instead of ! -z. [SC2236]	395:8
Info	Double quote to prevent globbing and word splitting. [SC2086]	402:15
Info	Check exit code directly with e.g. 'if mycmd;', not indirectly with $?. [SC2181]	403:9
Error	> is for string comparisons. Use -gt instead. [SC2071]	462:26
Info	Double quote to prevent globbing and word splitting. [SC2086]	484:16
```

### DIFF
--- a/webodm.sh
+++ b/webodm.sh
@@ -18,7 +18,6 @@ if [[ $platform = "Windows" ]]; then
 	export COMPOSE_CONVERT_WINDOWS_PATHS=1
 fi
 
-default_nodes=1
 dev_mode=false
 gpu=false
 
@@ -37,6 +36,7 @@ DEFAULT_MEDIA_DIR="$WO_MEDIA_DIR"
 DEFAULT_SSL="$WO_SSL"
 DEFAULT_SSL_INSECURE_PORT_REDIRECT="$WO_SSL_INSECURE_PORT_REDIRECT"
 DEFAULT_BROKER="$WO_BROKER"
+DEFAULT_NODES="$WO_DEFAULT_NODES"
 
 # Parse args for overrides
 POSITIONAL=()
@@ -103,7 +103,6 @@ case $key in
     shift # past value
     ;;
     --no-default-node)
-    default_nodes=0
     echo "ATTENTION: --no-default-node is deprecated. Use --default-nodes instead."
     export WO_DEFAULT_NODES=0
     shift # past argument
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @gromain looks great! :pray:

PR #1137, opened 2022-02-01
Fix issue with plugins elevationmap and changedetection

gromain:
This closes #1058 and #1059.

More tests are needed to make sure nothing breaks. Also, image size has increased by about 400MB. Not sure if there is anything we can do on that.

PR #1139, opened 2022-02-02
Fix issue with plugins elevationmap and changedetection _____ 2

rumenmitrev:
in reletaion to:
https://github.com/OpenDroneMap/WebODM/pull/1137
image size is + ~200MB
But we/plugins need desperatly those tools

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -28,7 +28,7 @@ jobs:
       id: docker_meta
       uses: crazy-max/ghaction-docker-meta@v1
       with:
-        images: opendronemap/webodm_webapp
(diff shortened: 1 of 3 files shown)


pierotofy:
Thanks for the PR, but we are not going to add these dependencies to make the plugins work; the current problem with the plugins is being discussed in #1137. Feel free to have a fork that installs the dependencies, but it's not good for users that don't use these plugins to have an extra ~200MB of downloads. :pray:

rumenmitrev:
You have point.
I’ve done it to myself already.
But I thought that some users want to use those plugins and not able to.
And 200 MB is nothing.
opendronemap/nodeodm:gpu is 2.6 G
Never mind.

PR #1157, opened 2022-03-17
Update sidebar.html

syxs:
Adds meters/feet/inch as display lengths in the UI under Scene sidebar


pierotofy (approved):
Excellent, thank you!!

PR #1159, opened 2022-03-25
Swagger and Django,DRF version

OKaluza:
Letting you know just in case you want to use this.
We wanted to use swagger openapi docs so I installed drf-yasg. (See here for example: https://asdc.cloud.edu.au/swagger https://asdc.cloud.edu.au/redoc )

This required bringing the Django and DRF versions forward a little so these modifications might save you some time if you were planning on updating them at some point.

If you don't want the swagger stuff, but the package upgrades are useful I can also make a pull request with just that commit.


pierotofy (approved):
Looks great @OKaluza :pray: thanks for taking the time to upgrade Django & co. The swagger/redoc functionality is pretty cool. :+1:

Issue #1163, opened 2022-03-29
The Edit Task Options dialog is cut off when called from the cloudimport or dronedb import plugins


OKaluza:
### How did you install WebODM (docker, installer, etc.)?
Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)
Firefox 98 on Ubuntu Linux

### What is the problem?
The Edit Task Options dialog is cut off when called from the cloudimport or dronedb import plugins

### How can we reproduce this?
- Enable cloudimport or dronedb import plugins and set up username/pw etc
- Import using the plugin buttons
- Use the Edit button to open the Edit Task Options dialog
- The dialog is cut off when opened from a modal dialog and very hard to work with
![image](https://user-images.githubusercontent.com/6336397/160539658-394d1d17-8b41-4d65-926c-4dd6bdcc1f4a.png)

I have implemented a fix and will reference this issue in the pull request

pierotofy:
Thanks for the detailed report (and the fix!). This can be closed now. :+1:

PR #1164, opened 2022-03-29
Fix for Edit Task Options dialog cut off when opened from plugin

OKaluza:
The Edit Task Options dialog is very small if opened from the cloud
import or dronedb import plugin modal dialogs, this css tweak fixes it
to take up the full height of the window.


OKaluza:
Fix for #1163, after applying this the dialog will take up the full window height as in this screenshot (still a bit odd at the bottom but much more usable): 
image

pierotofy (approved):
Looks great, thank you! :pray:

PR #1197, opened 2022-06-10
Add option to set trusted hostname for db connections

OKaluza:
For improved security, it's useful to be able to limit which host can
connect to the database (by setting POSTGRES_ALLOW_HOST to ip or name
instead of the default 'all'


pierotofy (approved):
Thanks @OKaluza ! Looks great.

Issue #1196, opened 2022-06-09
Possible Bug: View Map (2D Map) not available when Orthophoto skipped and DEMs Enabled


Saijin-Naib:
### How did you install WebODM (docker, installer, etc.)?
Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)
Microsoft Edge, Alpine Linux Edge (v3.16.x)

### What is the problem?
When you skip the generation of an orthophoto but still opt to generate DEM products (DSM/DTM), the View Map (2D Map) function is not shown in the Task list, despite being valid.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)
Generate a Task with skipping of Orthophoto, but generation of DEM products.
Observe View Map button being missing.
![image](https://user-images.githubusercontent.com/19295950/172750526-b012e6b0-b7e2-411c-9930-210a0f219f19.png)

### Further Information
I would suggest that even if we skip orthophoto and skip all DEM product generation, that we should _still_ have the View Map function so we can see the locations of the calibrated cameras, as that is salient information for the pointcloud/mesh/etc.

PR #1205, opened 2022-07-08
Smart Tiles, Faster Tiles

pierotofy:
This PR adds:
 - [x] "Smart" tiles with mixed PNG/JPEG format (similar to marble cutter), wherein tiles that have no transparency are rendered to JPG (smaller download size, faster) and edge tiles with transparency use the heavier PNG
 - [x] 512x512px tile sizes (upped from 256) as to reduce the number of network requests and render more map space in fewer network calls
 - [x] Improved the hillshade rendering of DSM/DTMs by ~25-30% by avoiding the multiple calls to `tile` and using `tile_buffer` instead.  
 - [x] Speeds up significantly DSM/DTM rendering when hillshading is not needed.


pierotofy:
- [x] WEBP support for transparent tiles when possible instead of PNG

pierotofy:
Fixes #1196

PR #1210, opened 2022-07-24
Use Ubuntu 20.04

pierotofy:
Upgrading to 22.04 will require more effort.﻿


pierotofy:
Nvm, lacks the latest gdal.

PR #1220, opened 2022-08-09
Bump mapping, environment maps during texture render

pierotofy:
Adds support for bump mapping and env maps for textured model rendering.
﻿


pierotofy:
anim

anim

Saijin-Naib:
We're slowly converging on being a sourceport for Quake :rofl: 

This looks incredible. It enhances legibility of the scene objects to an incredible level.

PR #1233, opened 2022-09-07
Losslessly compress test images

Saijin-Naib:
About 8% total filesize reduction.

Lossy, or alternative formats (webp, AVIF, HEIF) possible in a follow-up pending discussion


pierotofy:
I'm hesitant to commit this since it will increase the size of the repo (despite the 8% savings). :grimacing: 

These images probably should have never been added to git in the first place (and that was my fault).

PR #1240, opened 2022-09-20
FIX missing GPU in wsl

### LINKED ISSUE #1239: WSL can't find GPU, since you can't see it with "lspci"
### How did you install WebODM (docker, installer, etc.)?

WSL

Chrome

### What is the problem?
On WSL you won't find gpu.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

In the scrip for the webodm.sh, you can't find the gpu with lspci, at least on my machine it wasn't found.
```
if lspci | grep 'NVIDIA'; then
	echo "GPU_NVIDIA has been found"
	export GPU_NVIDIA=true
	set -e
	return
fi

if lspci | grep "VGA.*NVIDIA"; then
	echo "GPU_NVIDIA has been found"
	export GPU_NVIDIA=true
	set -e
	return
fi
```
I have removed the if statements and the gpu works like a charm.  My suggestion is to use the code below as an extra (3rd one) , since I usually use it.

```
NVIDIA_DEVICE=$(nvidia-smi -L | grep "GPU 0:" | awk -F ': ' '{print $2}' | awk -F '(' '{print $1}')
#IF the NVIDIA_DEVICE has the word NVIDIA in it, then enable NVIDIA
if [[ $NVIDIA_DEVICE == *"NVIDIA"* ]]; then
	echo "GPU_NVIDIA has been found"
	export GPU_NVIDIA=true
	set -e
	return
fi
```

IuliuNovac:
FIX #1239

### DIFF
--- a/webodm.sh
+++ b/webodm.sh
@@ -185,6 +185,14 @@ detect_gpus(){
 			set -e
 			return
 		fi
+		NVIDIA_DEVICE=$(nvidia-smi -L | grep "GPU 0:" | awk -F ': ' '{print $2}' | awk -F '(' '{print $1}')
+    #IF the NVIDIA_DEVICE has the word NVIDIA in it, then enable NVIDIA
+    if [[ $NVIDIA_DEVICE == *"NVIDIA"* ]]; then
+    	echo "GPU_NVIDIA has been found"
+    	export GPU_NVIDIA=true
+    	set -e
+    	return
+    fi
 
 		if lspci | grep "VGA.*Intel"; then
 			echo "GPU_INTEL has been found"


IuliuNovac:
Tested on a machine with an Nvidia GPU, it must be run on one without. I use this method for other scrips, so I am pretty sure it works on pc with integrated GPU(no Nvidia GPU), but for pc with multiple GPUs where the GPU 1-X is not Nvidia will still miss it. On a laptop, with integrated GPU and Nvidia GPU, it works without issues.

pierotofy:
Excellent, thanks @IuliuNovac ! :pray:

PR #1246, opened 2022-10-01
Plant Health Enhancements

pierotofy:
- Allow min/max input freedom
- Add -1,1 range to ratio formulas


syazoo:
Great job. I love that the jobs I had already processed worked perfectly with this index value range.

PR #1247, opened 2022-10-06
Updated to check if a public link to use the organization website instead of /

ssantacroce-uwach:
If I am using this with a client, I don't want them to be able to click on my logo and go to my public dashboard. Not like they could login, but I feel like it should link to the organization website since they can't do anything in the backend (the point of the public link).

This is my first PR with WebODM. In addition, I am not an expert in Django, so if there's an easier or more "Django-y" way to do this, please let me know. Thank you!

### DIFF
--- a/app/templates/app/base.html
+++ b/app/templates/app/base.html
@@ -93,7 +93,7 @@
                 </button>
                 {% block navbar-top-links %}{% endblock %}
             
-                {% if '/public/task/' in request.path and SETTINGS.organization_website %}
+                {% if request.user.is_anonymous and SETTINGS.organization_website %}
                     <a class="navbar-brand" href="{{ SETTINGS.organization_website }}"><img src="{% settings_image_url 'app_logo_36' %}" alt="{{ SETTINGS.app_name }}" /></a>
                     <a class="navbar-link" href="{{ SETTINGS.organization_website }}"><p class="navbar-text">{{ SETTINGS.app_name }}</p></a>
                 {% else %}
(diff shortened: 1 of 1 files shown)


pierotofy:
Hey @ssantacroce-uwach :wave: thanks for the PR! I unfortunately cannot merge this modification, as not everyone who uses WebODM also sets, or even has, an organization website (it's blank by default). The way to do this might be to keep this modification on your fork and rebuild/update WebODM when needed.

ssantacroce-uwach:
So I could be mistaken, but with a new system the organization_website gets set to either the WebODM or OpenDroneMaps page. It is not blank. 

But if that is a concern, would a check that the organization_website not being blank suffice?

pierotofy:
Ah, good point; it's actually set to `https://github.com/OpenDroneMap/WebODM/`. 

I guess a good way to do this would be to move the default value to a variable, then expose that variable to the django view (or make a django function) and check to see if the value has been changed from the default.

ssantacroce-uwach:
I just modified it to go to '/' if the org url is blank. It really shouldn't matter if the url is the default. If you're on the "public" link, it shouldn't have a url that goes back to the dashboard, because that isn't something a public viewer would be able to login to. And if it's a logged in user for whatever reason viewing the public link, they know how to get back to the dashboard...

pierotofy:
I got it; how about checking if the user is logged in?

 * Public URL and anonymous? --> org website (or "/" if blank)
 * Else --> "/"

ssantacroce-uwach:
That is a way better idea! Sorry I'm very new to Django, my main framework is FastAPI. Your suggestion has been implemented!

pierotofy:
Awesome, thanks!!

PR #1249, opened 2022-10-14
Start of the upgrade script and new docker-compose file

ssantacroce-uwach:
Ok so I didn't try running the actual script cuz it's missing a lot of steps still. But I did run each step manually and it dumped the db data, created the new databse, and loaded in the data to the new db. 

I still need help and direction on how we want to decommission the old database and switch to the new database automagically... suggestions are welcome!

### DIFF
--- a/docker-compose.db_14.yml
+++ b/docker-compose.db_14.yml
@@ -1,0 +1,14 @@
+version: '3.7'
+
+services:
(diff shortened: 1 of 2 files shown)


pierotofy:
Closing as this is still in draft state. :pray:

PR #1257, opened 2022-10-25
Fix the "output truncated at undefined lines" message

OKaluza:


pierotofy (approved):
Nice catch! Thanks @OKaluza

PR #1266, opened 2022-11-21
Update boot.py with some option tweaks

smathermather:
Could use some review from @Saijin-Naib as well... .

### DIFF
--- a/app/boot.py
+++ b/app/boot.py
@@ -101,14 +101,13 @@ def add_default_presets():
     try:
         Preset.objects.update_or_create(name='Multispectral', system=True,
                                         defaults={'options': [{'name': 'auto-boundary', 'value': True},
-                                                              {'name': 'radiometric-calibration', 'value': 'camera'},
-                                                              ]})
+                                                              {'name': 'radiometric-calibration', 'value': 'camera'}]})
         Preset.objects.update_or_create(name='Volume Analysis', system=True,
                                         defaults={'options': [{'name': 'auto-boundary', 'value': True},
                                                               {'name': 'dsm', 'value': True},
                                                               {'name': 'dem-resolution', 'value': '2'},
                                                               {'name': 'pc-quality', 'value': 'high'},
-                                                              {'name': 'use-3dmesh', 'value': True},]})
+                                                              {'name': 'use-3dmesh', 'value': True}]})
         Preset.objects.update_or_create(name='3D Model', system=True,
                                         defaults={'options': [{'name': 'auto-boundary', 'value': True},
                                                               {'name': 'mesh-octree-depth', 'value': "12"},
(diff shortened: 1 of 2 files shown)


Saijin-Naib:
Like it a lot!

My only possible point of concern would be what might folks be using Ultra Quality preset for, and should we set the floor of the GSD lower? 1.0cm/px? Maybe 0.1cm/px? If someone's doing ground-level photogrammetry or reconstruction, they can certainly be at that GSD or finer, and might be disappointed and bounce off if they get 2cm/px from us and whatever intrinsic GSD from another processing suite...

smathermather:
Yeah, I've been wondering how best to handle GSD defaults, since the estimate over-rules but won't if it's coarser.

Saijin-Naib:
I sort of feel like, for anything not "Default" we can probably push that really low, or at least match Lightning, and hopefully avoid a bunch of dissapoints. I've seen too many folks think our output is trash because we set a reasonably coarse GSD estimate for the processing profiles we ship.

But balancing that against folks not having the proper-sized hardware is really hard, you're right... Maybe since this is an Ultra preset we can sort of absolve ourselves of any responsibility to keep the barrier to entry low here and be okay?

pierotofy:
I'm a bit worried the settings will make processing fail for folks that don't have sufficient memory (for the Urban in particular, the new Ultra preset is probably OK, even though there's no disclaimer). Although the Ultra Quality preset might raise questions as to how it's different than the "High Resolution" preset.

smathermather:
> What's the end goal for the proposed changes?

The purpose is to solve the other issue, which is folks trying to figure out how to get the best possible results. We have two categories of complaints on the two ends of the spectrum: crashes due to settings and complaints of quality. The most common response I make on the forums to improve quality is this preset. I do it maybe a dozen times a month. If it were a preset, folks would probably try it when they wanted improved quality.

Happy to drop it for the building preset though or splitting off to a separate building preset.

pierotofy:
Ok!

I think once this is fixed we can merge.

```
  File "/webodm/app/boot.py", line 158
    defaults={'options': [{'name': 'auto-boundary', 'value': True},
    ^
SyntaxError: expression cannot contain assignment, perhaps you meant "=="?
```

smathermather:
I think it's fixed (there were existing inconsistencies that I patched as well as the outright errors I introduced...). But the builds always fail for me. Did a quick syntax check with `python -m py_compile boot.py`.

pierotofy:
Ah, you probably need to update the unit test on `app/tests/test_api_preset.py` (change `14` to `15`) since a new preset was added.

```
File "/webodm/app/tests/test_api_preset.py", line 61, in test_preset
    self.assertTrue(len(res.data) == 14)
AssertionError: False is not true
```

PR #1275, opened 2022-12-10
Silence django's session data corrupted warnings

pierotofy:
As the title says. This warning has been confusing users for years.


Saijin-Naib:
image

PR #1286, opened 2023-01-28
Binary glTF support

pierotofy:
This PR adds support for displaying draco compressed glTF textured models, which are both smaller and faster. Also removes the COLLADA importer code for showing camera locations (using glTF now).

Also adds downloads of glTF textured model assets when they are available.


pierotofy:
Ref https://github.com/OpenDroneMap/ODM/pull/1587

PR #1291, opened 2023-02-21
Cleanup Help Links On Diagnostics Page

Saijin-Naib:
Replace MacOS link to StackExchange with direct link to Docker docs, improve deep-link of Windows Docker documentation, add link to WSL2 documentation, change Windows display to support both Hyper-V and WSL2 (with appropriate links).

### DIFF
--- a/coreplugins/diagnostic/templates/diagnostic.html
+++ b/coreplugins/diagnostic/templates/diagnostic.html
@@ -27,10 +27,10 @@
         {% endif %}
     </div>
 </div>
-
+git
 <hr/>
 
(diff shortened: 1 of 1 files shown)


coreplugins/diagnostic/templates/diagnostic.html
```diff
@@ -27,10 +27,10 @@ <h4>{% trans 'Memory' context 'Computer memory (RAM)' %}</h4>
         {% endif %}
     </div>
 </div>
-
+git
```
pierotofy, line 30: Should this string be here?

Saijin-Naib: No, absolutely not. Thank you!

PR #1293, opened 2023-02-22
Add links for Linux Engine and Desktop

Saijin-Naib:
Add links and verbiage for Docker Desktop For Linux and Docker Engine resource tuning/allocation.

This maintains consistency with the other platforms, and it helps to surface documentation (for Engine, specifically) that I felt took me too long to find from the first-party in terms of setting runtime and other resource limits under Docker Engine.


pierotofy:
Thanks Brett, although I think we might be getting a bit too specific; the diagnostic page should not contain a guide on using docker?

Saijin-Naib:
> Thanks Brett, although I think we might be getting a bit too specific; the diagnostic page should not contain a guide on using docker?
No, indeed not. My goal was to make sure that each platform was present with a link directly to the associated documentation from Docker (or the appropriate first-party) on how to tune resource allocation. Linux, like Windows, happens to have two different methods of using Docker with two different means of tuning resource allocations.

Is the verbiage of how I talk about Linux Docker too wordy? Would it be better to just have it more like Windows?

pierotofy:
I think ideally we would just:
 1. Improve the docs on docs.opendronemap.org (if we don't have.a page on docker resource allocation already)
 2. Link the page on the docs (single link).

Saijin-Naib:
My thought on that is doesn't that make us responsible for making sure if anything changes on Docker's side we keep our version of the docs in sync, whereas if we link upstream then that is no longer our thing to maintain?

I'm open to whatever strikes you best, however.

pierotofy:
We should have a page in the docs, and link to the docs.

Saijin-Naib:
Awesome, I'll do that.

How do you feel about a general link to the top level doc on resource allocation versus a deep link per platform to our docs (how it is in this PR, just redirected)?

pierotofy:
> a general link to the top level doc on resource allocation

:+1:

pierotofy:
Closing and waiting for a new PR as discussed. :+1:

PR #1297, opened 2023-03-07
Better Project and Task Management

pierotofy:
Implements project tags, task tags, sorting, filtering and search functionalities for saner project management.

Work in progress.

image


Saijin-Naib:
Is arbitrary drag/drop or other manual sort in this work? I'm not understanding the code well enough to tell.

This looks amazing, though 😎

pierotofy:
Not quite; but it does allow arbitrary sorting via tags.

pierotofy:
Brief summary of changes:
 * Added sorting at project level / task level. The sort by owner is only available at the project level, because by definition in WebODM every task in a project belongs to the project's owner (so would make no sense to sort by owner at the task level). 
 * Added the ability to add tags at both task and project level.
 * Added search capabilities. On the backend, this functionality is available via the `/api/projects/?search=` endpoint, where `search` is a string that accepts: `name #projecttags ##tasktags`. (`name` --> string or strings to match project names or task names, `#projecttags` --> tags to match tags at project level, `##tasktags` --> tags to match tags at the task level). Multiple terms can be combined. Tags that begin with a dot (`.`) are hidden to the WebODM UI, but are internally tracked and will respond to search queries. Hidden tags cannot be added via the WebODM UI and are meant for programmatic use only.
 * Added/expended sort capabilities via `/api/projects/?ordering=[-]fieldname`. (optional minus sign indicates descending order).
 * Filtering at the task level is done client-side.

PR #1320, opened 2023-03-31
Update Formulas.py - Include the MPRI index

vagner-silveira:
I would like to suggest the inclusion of the MPRI vegetative index (Modified Photochemical Reflectance Index), as it has a 90% correlation with the NDVI in studies carried out for use in large cultures here in Brazil, with RGB cameras. the code would be just below the VARI index on lines 47 to 51.
the index vegetative '(G - R) / (G + R)'


pierotofy:
Thanks @vagner-silveira looks great!

PR #1321, opened 2023-04-02
Adding a --db-dir option to specify an external Postgres data dir

tariqislam:
Feature:
Similar to the --media-dir option, I thought it would be useful to add a --db-dir option.

Use case:
If you are in AWS, and you have different EBS volumes for --media-dir and --db-dir, when you restack your WebODM EC2 instance you can reattach these persisting EBS volumes to the new EC2 instance, hence avoiding data migration or the fact that data exists in docker volumes.

This feature is completely backward compatible and transparent for existing functionality.

Tests:
Seems like the webapp tests passed:

Test Suites: 38 passed, 38 total
Tests:       39 passed, 39 total
Snapshots:   0 total
Time:        1.228s, estimated 2s
Ran all test suites.

However, the tests fail when trying to run the NodeODM tests. I'm not sure how to make those work. Regardless, this feature is at a container level and does not affect the NodeODM container.

In addition, I manually tested that with the default (docker volume) and the --db-dir options work.


tariqislam:
@pierotofy @Saijin-Naib I created a small feature to add --db-dir option. Please review the code. Thanks.

pierotofy:
Thanks @tariqislam it looks good, but can you please revert the commit that removes the welcome messages?

tariqislam:
> Thanks @tariqislam it looks good, but can you please revert the commit that removes the welcome messages?

Sorry, missed that. It's been reverted.

PR #1323, opened 2023-04-03
Feature add http read timeout setting

tariqislam:
When importing a task, the default timeout is 10 seconds. This is too short for large imports, and the timeout value should be config-driven and modifiable through the command line. Hence adding that feature.


tariqislam:
Another PR @pierotofy @Saijin-Naib  :)

tariqislam:
Apologies for the few commits... tested as working.

pierotofy:
The timeout value indicates the maximum amount of time that the server has to *begin* sending the requested file (it doesn't matter how big the files is).

If 10 seconds aren't enough I'm fine bumping it to a higher value, but how did you encounter issues with the current value?

tariqislam:
> The timeout value indicates the maximum amount of time that the server has to _begin_ sending the requested file (it doesn't matter how big the files is).
> 
> If 10 seconds aren't enough I'm fine bumping it to a higher value (maybe 20-30), but how did you encounter issues with the current value?

I was iusing the "Import from URL" functionality from one WebODM to another -- specifically, the source was a URL to a task and not a file upload. I was facing errors when importing any project that wouldn't send the data from one webodm to another in less than 10 seconds.

tariqislam:
I've also added a possible bug https://github.com/OpenDroneMap/WebODM/issues/1324 when importing a large asset as a task.

pierotofy:
> I have seen a large webodm project take a long time (minutes)

This is probably the part that should be fixed, rather than increasing the timeout. It's strange that you're seeing such very long times though, WebODM uses zipfly (https://github.com/OpenDroneMap/WebODM/blob/master/app/models/task.py#L459) to immediately start streaming a download, so the size of the files should not stop the download from beginning quickly even with large files.

Where are you mounting the app/media folder from? If you're using anything other than a normal mount to a normal disk you might have delays that are not due to WebODM.

tariqislam:
> > I have seen a large webodm project take a long time (minutes)
> 
> This is probably the part that should be fixed, rather than increasing the timeout. It's strange that you're seeing such very long times though, WebODM uses zipfly (https://github.com/OpenDroneMap/WebODM/blob/master/app/models/task.py#L459) to immediately start streaming a download, so the size of the files should not stop the download from beginning quickly even with large files.
> 
> Where are you mounting the app/media folder from? If you're using anything other than a normal mount to a normal disk you might have delays that are not due to WebODM.

Agreed, if WebODM starts streaming the zip file immediately, that makes sense for the fix to be in the streaming side rather than a larger timeout to start the streaming.

The machine doing this is a standalone Ubuntu 20 box. WebODM is running in docker, but is mounting a local file system folder (/var/dev) as the media folder. 

At the hardware level, the /var/dev folder is actually a mount point to a RAID 10 setup, again local to the machine. There is a RAID card, with 4 SSD drives attached. This machine acts as a NAS for my project storage, and the RAID setup is the NAS folder as well. I have copied files and read files directly to/from the NAS folder from other machines, and have never seen any slowdown at all.

pierotofy:
So the next thing to do is figure out, in this particular hardware setup (which I don't have) why WebODM/zipfly takes longer than usual to begin a zip download.

tariqislam:
@pierotofy I tested this on another system, (Laptop, 64GB RAM, Ryzen 5, one SSD), and I have verified that the zip file does not start streaming immediately. Sometimes it waits for upwards of a minute before starting to stream. So I believe there is a bug somewhere in the functionality that starts the streaming.

So in the mean time, this PR would be relevant as a fix for the above issue.

tariqislam:
@Saijin-Naib @pierotofy Please let me know what your latest thoughts are on this. I'm not sure how many people are using the functionality of uploading assets from one WebODM to another via API (a great feature btw) but the issue with the zipping mechanism would affect such a transfer. If it would take time to find the root cause of the zipping issue, this PR is a fix to go around it. Would be great to merge it.

Saijin-Naib:
I'm not sure either, but we very infrequently get reports of failure to sideload data using the Import function from the UI, so those cases might have been related, though impossible to tell now.

pierotofy:
Increasing the timeout is not a proper fix, imo.

> the next thing to do is figure out, in this particular hardware setup (which I don't have), why WebODM/zipfly takes longer than usual to begin a zip download.

We'd welcome help to investigate this further.

Issue #1326, opened 2023-04-15
WARNING Session data corrupted


ichsan2895:
**How did you install WebODM (docker, installer, etc.)?**
With Docker
```
$ docker --version
Docker version 20.10.8, build 3967b7d
```

**What's your browser and operating system?**
Opera 78 on Linux

**What is the problem?**
WARNING Session data corrupted

**How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing?**

This is happen after I updated to newest WebODM

```
# download newest git then extracted it to a folder named "WebODM-master"
$ cd WebODM-master

# update WebODM with GPU supports
$ ./webodm.sh update --gpu
01:00.0 3D controller: NVIDIA Corporation GP108M [GeForce MX150] (rev a1)
GPU_NVIDIA has been found
Checking for docker...   OK
Checking for docker-compose...  OK
docker-compose -f docker-compose.yml -f docker-compose.nodeodm.gpu.nvidia.yml -f docker-compose.nodemicmac.yml down --remove-orphans
Removing webapp                   ... done
Removing worker                   ... done
Removing webodm-master_node-odm_1 ... done
Removing db                       ... done
Removing broker                   ... done
Removing network webodm-master_default
Updating WebODM...
Skipping source update (.git directory not found)
docker-compose -f docker-compose.yml -f docker-compose.nodeodm.gpu.nvidia.yml pull
Pulling broker   ... done
Pulling db       ... done
Pulling worker   ... done
Pulling node-odm ... done
Pulling webapp   ... done
Done! You can now start WebODM by running ./webodm.sh start

# start it with GPU
$ ./webodm.sh start --gpu
01:00.0 3D controller: NVIDIA Corporation GP108M [GeForce MX150] (rev a1)
GPU_NVIDIA has been found
Checking for docker...   OK
Checking for docker-compose...  OK
Starting WebODM...

Using the following environment:
================================
Host: localhost
Port: 8000
Media directory: appmedia
Postgres DB directory: dbdata
SSL: NO
SSL key: 
SSL certificate: 
SSL insecure port redirect: 80
Celery Broker: redis://broker
Default Nodes: 1
================================
Make sure to issue a ./webodm.sh down if you decide to change the environment.

docker-compose -f docker-compose.yml -f docker-compose.nodeodm.gpu.nvidia.yml up --scale node-odm=1
Creating broker                   ... done
Creating db                       ... done
Creating webodm-master_node-odm_1 ... done
[...]

PR #1327, opened 2023-04-16
Re-Silence Django Warning (Session Data Corrupted)

Saijin-Naib:
It appears the logging object we silenced was only for versions prior to v2.2.x, which we use now.


pierotofy:
Thanks @Saijin-Naib :+1: closes #1326

PR #1331, opened 2023-04-26
Columns are added to the projects view in administration.

diegoaces:
Hello @pierotofy , 
I made some changes in the administration view, I added a column that shows the number of tasks associated with the project, also a filter to search and other columns.
I hope it is useful.

image


pierotofy:
Thanks @diegoaces ! Looks great. 👍

Issue #1336, opened 2023-05-03
Docker-compose config publishes db and nodeodm ports on host


t4y:
### How did you install WebODM (docker, installer, etc.)?

I am running WebODM on Linux using Docker. I am not using the scripts but instead starting the setup with `docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up -d`.

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Firefox 112 on Ubuntu Linux

### What is the problem?

I noticed that running WebODM like this (didn't try to see if the scripts mitigate that) publishes postgresql, and the nodeodm web interface on the host under ephemeral ports because of the `ports:` syntax:

https://github.com/OpenDroneMap/WebODM/blob/48d76079bd04d618b4e83a8b414f033a5fd8bf62/docker-compose.nodeodm.yml#L14-L15

https://github.com/OpenDroneMap/WebODM/blob/48d76079bd04d618b4e83a8b414f033a5fd8bf62/docker-compose.yml#L11-L12

which opens ports on the host:
```
32768/tcp open  postgresql PostgreSQL DB 9.4.13 - 9.4.15
32769/tcp open  http       Node.js Express framework
|_http-cors: HEAD GET POST PUT DELETE PATCH
|_http-title: NodeODM - Web UI
```

Having some random ports unnecessarily be opened might be seen as undesirable especially if one prefers to keep everything behind a reverse proxy.

Using [expose:](https://docs.docker.com/compose/compose-file/compose-file-v3/#expose) instead of `ports:` fixes that for me.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

run `docker-compose -f docker-compose.yml -f docker-compose.nodeodm.yml up -d` and then for example `ss -lntup` to see that two new TCP ports have been opened on the host.

Saijin-Naib:
I'm not sure if this might be related to the change that just landed with the port mapping?:  
https://community.opendronemap.org/t/webodm-dockers-cant-reach-internet/15831

PR #1337, opened 2023-05-03
Docker-compose: expose ports instead of publishing

t4y:
With the
```
ports:
  - "12345"
```

syntax Docker publishes the ports under an ephemeral port on the host. Since these ports should presumably only be available to other services, using expose: instead avoids publishing unnecessary services. See #1336


pierotofy:
Thanks for the report and patch @t4y ! :pray:

PR #1338, opened 2023-05-03
ports --> expose

pierotofy:
Fixes #1337


pierotofy:
Ah, they beat me to it! :)

PR #1344, opened 2023-05-18
Add cpu usage chart

Kathenae:
## CPU usage chart
---

This pull requests add a cpu usage chart to the diagnostics page, allowing users to view the system wide cpu usage percentage and update the charts every 5 seconds.
Changes:
- added `psutil` to the main `requirements.txt` and `coreplugins/diagnostics/requirements.txt`.
- Modified and Refactored `coreplugins/diagnostics/plugin.py` to retrieve the system wide CPU usage and also exposed an api endpoint that returns all diagnostics information the plugin uses.
- Modified `coreplugins/diagnostic/templates/diagnostic.html` to include a donut chart that displays the CPU usage percent.
- Refactored existing code to support updating the charts every 5 seconds with information retrieved using the exposed api endpoint.

Note: The cpu usage stats is retrieved using the psutil library which is considered cross-platform but i personally have not tested this in other operating systems so not sure what the behavior would be there.

this PR is related to this #1341 issue


pierotofy:
Thanks @Kathenae , :pray:  I've made some small adjustments, but otherwise looks great.

image

I'll merge this by tonight.

Kathenae:
Thanks!

PR #1346, opened 2023-05-21
Bugfix: Task notification plugin fails to send email notifications in production mode

Kathenae:
##  Description
This PR addresses an issue with the task notification plugin that prevented it from sending email notifications as expected in production mode. Specifically, the plugin was able to send notifications when using the Send test email button on the configuration form of the webapp, but failed to do so when triggered via a signal on a processing node. Although not entirely certain, this behavior may be traced to the plugin's inability to read/write the .conf file when running on processing node instances.

### Solution
To fix the issue, this PR refactors the plugin to use the plugin DataStore instead of the .conf file. This approach ensures that the plugin's configuration is accessible across all instances, allowing it to send email notifications as expected in both dev and production mode.

### Changes
In terms of changes made, the load() and save() functions have been refactored to use the plugin DataStore to retrieve and save the configuration, respectively. This eliminates the need for the .conf file and ensures that the plugin works correctly

### Tests
To confirm the fix, I tested the plugin locally in both dev and production environments and verified that email notifications are sent when tasks finish processing and when tasks fail during processing.

### DIFF
--- a/coreplugins/tasknotification/config.py
+++ b/coreplugins/tasknotification/config.py
@@ -1,40 +1,39 @@
-import os
-import configparser
-
-script_dir = os.path.dirname(os.path.abspath(__file__))
-
 def load():
-  config = configparser.ConfigParser()
-  config.read(f"{script_dir}/.conf")
-  smtp_configuration = {
-    'smtp_server': config.get('SETTINGS', 'smtp_server', fallback=""),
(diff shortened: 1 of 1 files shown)


pierotofy:
Makes sense, thanks for the fix!

pete1019:
Is a new release needed for that? If so please do so.
Or is uninstalling and installing the addon again doing the trick? Thanks

PR #1350, opened 2023-06-03
addon: Projects and tasks chart

diegoaces:
This pull request adds the ability to see the number of projects and tasks per month done in webodm through a chart.

image


pierotofy:
Great idea for a plugin, thanks @diegoaces !

PR #1353, opened 2023-06-18
Fixed the uploaded file was too large, causing 10 retries

vinsonliux:
report：
https://community.opendronemap.org/t/cannot-upload-file-exceeded-max-retries-10/16360


pierotofy:
Thanks @vinsonliux 🙏 , a few things should be fixed before I can merge this:

 * Can we not change the formatting/indentation of parts that are not relevant to this fix?
 * Should the max file size limit be higher than 2048? Video files can be several GBs. Maybe 64/128 GBs?

PR #1354, opened 2023-06-18
Add delete button for read only projects

pierotofy:
Adds a delete button to be shown to users that have read-only 
permissions to a project, since they should be able to remove themselves 
from a project that was shared with them.

image


Saijin-Naib:
Should the verbiage be something more like Remove? I'm not sure if Delete implies something more permanent or affecting the source to me, but I'm struggling with what would be better and still be concise. Or I'm over thinking this, which is possible 🤷

pierotofy:
There's a dialog box that pops out informing the user that the project is not really destroyed when the link is pressed.

PR #1355, opened 2023-06-19
Dropzonem maxfilesize

vinsonliux:
Fixed the uploaded file was too large, set maxFilesize = 128G

TODO：
1. No video files larger than 10gb have been tested
2. Drag-and-drop upload can select any type of file,no accepted file will be retried 10 times


pierotofy:
Looks great, thank you @vinsonliux !

PR #1364, opened 2023-07-24
add "volumes" parameter in  docker-compose.nodeodm.yml to fix misunderstand 

dust27149:
Nodeodm consumes a lot of space on system disk , and many users misunderstand the effect of parameter "--media-dir".In many opinion,the temporary files and results should be saved in the same floder,which can be modified by "--media-dir",but it dosen't.Add "volumes" parameter in "docker-compose.nodeodm.yml",so that the temporary files produced during the construction will be saved in the same floder as passed in "--media-dir".


pierotofy:
Hey @dust27149 :wave: thanks for the PR!

The main issue with NodeODM using a mounted folder is that (on Windows only) there used to be (there still is?) a quirky bug which caused files to become corrupted once in a rare while.

There might also be performance differences on Windows/macOS.

I'm OK adding support for storing processing files in a volume, but should be optional and turned off by default. Perhaps the PR could be adjusted to reflect this?

dust27149:
I have test the code on Ubuntu20.04,it works.It remains to be seen whether it works on Windows/MacOS.

dust27149:
Something to discuss. The following solution is common to see when depolying WebODM+NodeODM+ClusterODM.
微信图片_20230726154932
ClusterODM is often used to link together multiple NodeODM API compatible nodes under a single network address. However,it can not access to the "webodm_node-odm_1" which is launched by WebODM. We can pass "--default-nodes 0" when starting webodm.sh and run another NodeODM in docker,but it seems like "bring sand to the beach". Shall we add the following config in "docker-compose.nodeodm.yml" ?
```yml
"ports:
      - "3000:3000"
```

pierotofy:
Probably not, because the default node is not secured by an access token.

dust27149:
Ok，thanks for your reply.

PR #1375, opened 2023-09-02
Update Map.jsx to use correct tile.osm.org URL

Firefishy:
See: https://github.com/openstreetmap/operations/issues/737


pierotofy:
Thanks @Firefishy !

Issue #1128, opened 2022-01-26
Database vacuuming either not running or unoptimized.


smathermather:
****************************************
### How did you install WebODM (docker, installer, etc.)?

Seriously dockerized here.

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Not relevant.

### What is the problem?

Autovacuum is either not set or untuned in WebODM

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

See this: https://community.opendronemap.org/t/what-is-eating-up-my-hard-disk-part-dieux-pssst-spoiler-its-the-database-not-being-vacuumed/10330

smathermather:
I'm guessing we touch autovacuum_vacuum_cost_limit and or autovacuum_vacuum_cost_delay in https://github.com/OpenDroneMap/WebODM/blob/master/db/Dockerfile maybe around line 58.

pierotofy:
I didn't even know this was a thing (why doesn't postgres do this automatically?), so I'll go with whatever suggestion.

smathermather:
> why doesn't postgres do this automatically

It's that old addage: _configuration over convention_. Wait... .

> https://github.com/OpenDroneMap/WebODM/blob/master/db/Dockerfile maybe around line 58.

If this is the right place, I'll take a stab at a pull request.

Saijin-Naib:
I'm going to defer to you, Stephen. I don't understand PGSQL enough to do anything other than guess. My familiarity is just with GeoPackage/SQLite on smaller deployments than we'd see ODM scale to.

OKaluza:
Hi, I've just spent a lot of time looking in to this issue and what causes it, it's hard to pin down, I've dug into the internals of postgresql a lot more than I wanted to and I have heaps of debugging info and further details I can share, but here are my conclusions so far:

- The autovacuum is running, and the settings should be ok - may be able to tweak them to get around this but I think it's better to fix the underlying cause
- The bloat is caused in a pg_toast_#### table related to app_tasks - this stores modified tuples from old records when a record is updated, the vacuum should clean them up, but clearly this isn't always happening:
```
SELECT                                           
       relname AS "table_name",
       pg_size_pretty(pg_table_size(C.oid)) AS "table_size"
FROM                            
       pg_class C
[...]

PR #1386, opened 2023-09-11
Move task.console_output

pierotofy:
Following up on the work done around #1128, this PR moves the console output storage to the filesystem to avoid Postgres vacuuming issues (and improves overall performance, since we no longer need to constantly update table rows every few seconds).


pierotofy:
FYI @OKaluza

pierotofy:
@smathermather might also be happy about this for larger tasks. :)

Saijin-Naib:
Oh, this is huge!

smathermather:
Oh. Yeah. Woohoo!

Issue #1394, opened 2023-09-15
Add ability to preserve secret_key between updates


pierotofy:
This would avoid forced logouts during software update.

PR #1395, opened 2023-09-16
Processing node handling improvements

pierotofy:
Adds settings to limit the number of nodes to display in the UI, adds more options to improve the update_node_info logic.


pierotofy:
Fixes #1394

PR #1396, opened 2023-09-16
Making the search easier

italogrmn:
Makes searching easier by searching for substrings in project names. Previously, it only found if the search string was identical to the project name

### DIFF
--- a/app/api/projects.py
+++ b/app/api/projects.py
@@ -68,7 +68,7 @@ class ProjectFilter(filters.FilterSet):
             project_name_vec = SearchVector("name")
             task_name_vec = SearchVector(StringAgg("task__name", delimiter=' '))
             name_query = SearchQuery(names, search_type="plain")
-            qs = qs.annotate(n_search=project_name_vec + task_name_vec).filter(n_search=name_query)
+            qs = qs.filter(name__icontains=names)
 
         if len(task_tags) > 0:
             task_tags_vec = SearchVector("task__tags")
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the PR @italogermano, however this would break search for task names within a project as well as removing the vector indexing (which is bad for performance).

italogermano:
@pierotofy I got the result I expected before by exchanging:

        if len(names) > 0:
            project_name_vec = SearchVector("name")
            task_name_vec = SearchVector(StringAgg("task__name", delimiter=' '))
            name_query = SearchQuery(names, search_type="plain")
            qs = qs.annotate(n_search=project_name_vec + task_name_vec).filter(n_search=name_query
            
Per:

        if len(names) > 0:
            project_name_vec = SearchVector("name")
            task_name_vec = SearchVector(StringAgg("task__name", delimiter=' '))
            qs = qs.annotate(n_search=project_name_vec + task_name_vec).filter(n_search__icontains=names)

pierotofy:
I understand that you're trying to make the search work for partial matches (e.g. have "te" match "test"). But I'm not sure this is what we want (even though I can see the use case).

The use of `__icontains` matches substrings, but has some shortcomings. For example:

Project name `test abc`: search query `abc test` --> no match
Project name `test`: search query `st` --> match. Arguably, a partial match could be helpful if it matched `te`, but not `st`?

https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/search/

italogermano:
If I understand, the problem should be solved with:

            name_query = SearchQuery(names, search_type="plain")
            qs = qs.annotate(n_search=project_name_vec + task_name_vec).filter(Q(n_search__icontains=names) | Q(n_search=name_query))

pierotofy:
Almost; 

the take-away is: let's not use `icontains`. :pray:

Issue #1244, opened 2022-09-30
Idle WebODM system's redis and celery processes eventually consume 100% CPU


robby-dermody:
### How did you install WebODM (docker, installer, etc.)?

From the instructions at https://github.com/OpenDroneMap/WebODM#getting-started

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Issue does not involve the client or browser. Server OS is Ubuntu 22.04.1 LTS as a VM running under VMWare ESXi (given 6 vCPUs, 32gb of RAM, 100GB of HD space). Running the newest `master` from git.

`free -m` output:
```
local@dronemap:~/WebODM$ free -m
               total        used        free      shared  buff/cache   available
Mem:           32105        2255       24589          15        5259       29382
Swap:           8191           0        8191
```

`df -h` output:
```
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.2G  1.7M  3.2G   1% /run
/dev/sda2        98G   28G   66G  30% /
tmpfs            16G     0   16G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.2G  4.0K  3.2G   1% /run/user/1000
```

`uname -a` output: `Linux dronemap 5.15.0-46-generic #49-Ubuntu SMP Thu Aug 4 18:03:25 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

### What is the problem?

I have a WebODM system that is totally idle (haven't processed a single thing on it yet). After starting up the services via the standard `./webodm.sh start` command and letting them sit for awhile (unsure exactly how long... a few hours, a few days), eventually the redis and celery processes will both consume 50-100% CPU (`uptime` shows `load average: 1.30, 1.33, 1.30`).

I looked at the logs of all docker processes, the ones that stick out are an infinite loop of the following from the `webapp` container:
```
redis.exceptions.ConnectionError: Error 104 while writing to socket. Connection reset by peer.
Exception ignored in: <function AsyncResult.__del__ at 0x7f9a50200f70>
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/celery/result.py", line 389, in __del__
    self.backend.remove_pending_result(self)
  File "/usr/local/lib/python3.9/dist-packages/celery/backends/asynchronous.py", line 180, in remove_pending_result
    self.on_result_fulfilled(result)
[...]

PR #1412, opened 2023-10-03
Adds support for automatically selecting the proper band filter

### LINKED ISSUE #1342: Automatically suggest the proper band filter
In multispectral orthophotos, the bands are labeled, so in theory it could be possible to let WebODM guess the proper band filter value instead of having the user manually select it.

pierotofy:
Adds automatic band order selection, which greatly simplifies the use of the Plant Health tab.

Up until now, users had the responsability of choosing the proper band order, which is tedius and error prone.

You can now just select "Automatic" (default) and let WebODM choose the proper band order for the algorithm of choice.

Closes #1342


pierotofy:
Also warns you if you don't have the proper bands for a selected algorithm:

image

pierotofy:
Should close #1244

PR #1415, opened 2023-10-08
Update README.md to use correct links to plugins directory.

allandaly:
Looks like the plugins directory changed from /plugins at some point to /coreplugins so this is a simple update to a few links in the readme.


Saijin-Naib (commented):
Wonderful, thanks for the cleanup!

PR #1417, opened 2023-10-17
Removed TGI plant health equation

rion-saeon:
I realised the TGI equation may be erroneous thus I removed it. I am surprised that no one questioned it before so hopefully it was not used much.

### DIFF
--- a/app/api/formulas.py
+++ b/app/api/formulas.py
@@ -54,10 +54,6 @@ algos = {
         'expr': '(2 * G) - (R + B)',
         'help': _('Excess Green Index (derived from only the RGB bands) emphasizes the greenness of leafy crops such as potatoes.')
     },
-    'TGI': {
-        'expr': '(G - 0.39) * (R - 0.61) * B',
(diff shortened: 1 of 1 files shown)


Saijin-Naib:
Oooh, interesting. Do you have any literature or examples?

Is our implementation flawed, or is the index itself flawed?

rion-saeon:
@Saijin-Naib the index I added are presumably flawed. [Here](https://www.researchgate.net/publication/340895640_Estimation_of_Triangular_Greenness_Index_for_Unknown_PeakWavelength_Sensitivity_of_CMOS-acquired_Crop_Images) is a correct version when camera wavelengths are unknown (a typical scenario for drone, kite etc. pilots). If someone could implement these equations (2-3 in this paper), it would be great.

rion-saeon:
I could not replicate the n-TGI equation with results. It may be safer to use this (which performs as expected): [A visible band index for remote sensing leaf chlorophyll content at the canopy scale](https://www.sciencedirect.com/science/article/abs/pii/S0303243412001791?via%3Dihub)

PR #1423, opened 2023-10-26
update app/boot.py

Saijin-Naib:
Add flag ```--texturing-skip-global-seam-leveling``` since I've seen a number of multispectral datasets processed without this and folks thinking our processing is broken due to getting out of range values.

### DIFF
--- a/app/boot.py
+++ b/app/boot.py
@@ -105,7 +105,8 @@ def add_default_presets():
     try:
         Preset.objects.update_or_create(name='Multispectral', system=True,
                                         defaults={'options': [{'name': 'auto-boundary', 'value': True},
-                                                              {'name': 'radiometric-calibration', 'value': 'camera'}]})
+                                                              {'name': 'radiometric-calibration', 'value': 'camera'},
+                                                              {'name': 'texturing-skip-global-seam-leveling', 'value': True}]})
         Preset.objects.update_or_create(name='Volume Analysis', system=True,
                                         defaults={'options': [{'name': 'auto-boundary', 'value': True},
                                                               {'name': 'dsm', 'value': True},


pierotofy:
This requires some thought, as in absence of proper metadata the radiometric calibration doesn't work, but skipping the global seam leveling will cause seam visibility.

Saijin-Naib:
Ugh... Multi-spectral is fun. 

What approach sounds reasonable?

An allow-list of known working sensors we pass it for?

smathermather:
I don't know what folks expect on the drone side, but on the satellite multispectral side, showing seams is an acceptable alternative when adequate reflectance corrections aren't available.

Saijin-Naib:
It seems that the sUAS expectation is that it "Just Works ™"

pierotofy:
I've received complaints when seams were present.

Saijin-Naib:
If returns count as complaints, it is non-zero for this...

smathermather:
Complaints that we weren't blending by default or complaints that there was a way to bypass blending? The latter would be a strange compliant, but I suppose we've seen stranger.

I'm surprised academic users haven't complained that we are blending without a method for bypassing. There should always be a way to bypass blending for scientific use cases.

But anyway, I have no skin in this game at this time, so 🤷.

Saijin-Naib:
Complaints that only our software has seams and no others do 🤷

I absolutely agree we should keep a way to not blend, but it appears more of our users expect it.

smathermather:
Random Postscript: super interested in the separation of illumination from surface color we see in neural applications. I would love to see incorporation of proper brdf (or modern equivalent) into in quantitative drone imaging.  No idea if anyone is working in this space, and I currently have no use cases myself at this time.

Saijin-Naib:
Thanks for the manual calculation of BRDF PTGISD. I had buried that 🙃

pierotofy:
I'll close this, as I think we should continue blending by default.

Issue #1431, opened 2023-11-10
Single-point measurement exporting bug


DanielALT0134:
Hello everyone,

I've been attempting to export some single-point measurements to a GeoJSON file on WebODM, but it's not working. Could you please help me with this?

Thanks in advance

pierotofy:
Should be fixed with #1435 thanks for the report!

Issue #1421, opened 2023-10-23
Wrong area measurement


Wilbodakwerken:
### How did you install WebODM (docker, installer, etc.)?

business installer

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

[[Chrome 118 on Windows 11](https://www.whatismybrowser.com/detect/what-version-of-chrome-do-i-have)

### What is the problem?

area measurement is not correct

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

When vertrical erea measurement is taken the oudput is not correct. 
For example i have a maesurement of 6,80m wide with a hight of 3,70m. This should give 25,16m². the software give me only 0,4m².

second example, a sloped roof thats 10,64m wide and 8,01 lenght it should be 85,23m², software tells me 77,2m²

Wilbodakwerken:
calculate the m² with the measurement and angels? 
below the screenshots, measurements are correct, m² isnt. 

![image](https://github.com/OpenDroneMap/WebODM/assets/148749278/5a9d173e-6fc0-42f1-a604-441e11cc7b3a)
![image](https://github.com/OpenDroneMap/WebODM/assets/148749278/e6dfe99a-7c48-43ea-9441-0c787ac28239)

pierotofy:
Should be fixed with #1435, thanks for the report! 🙏 

Note that it might take a few weeks before an update is released for Windows, so this fix will not be available right away.

PR #1435, opened 2023-11-14
10x volume calculations, base surface definitions. fix 3D vertical area measurements

pierotofy:
This PR speeds up volume calculations using the measure tool by an order of magnitude. It also adds support for more options for base surface definition:

 * Triangulate (*default*), tassellates the input point set and interpolates linearly
 * Plane, fits a 3D plane minimizing the distance of the plane to all points in the set
 * Average, uses the mean Z of all points
 * Highest, uses the highest Z of all points
 * Lowest, uses the lowest Z of all points

image

This PR also removes `grass-core` as a dependency, which was an obstacle to a possible future port to native macOS.

It moves the `changedetection` and `elevationmap` plugins to https://github.com/OpenDroneMap/WebODM-Plugins-Archive. I don't know who is using these plugins, but they have dependencies to the grass code that is now removed. It should be possible to update the plugins by using the more up-to-date `run_function_async` function (instead of relying on the previously used grass context/engine), but will require some work. The plugins are in need of a maintainer.


Saijin-Naib:
Also makes rolling this up for Alpine a bit easier for me 🤓

This is quite an exciting rework!

pierotofy:
Fixes #1431

pierotofy:
Fixes #1421

PR #1441, opened 2023-12-01
Notebook for optimizing DTM parameters

vonnonn:
This is a notebook that allows the user to iterate ground classifier parameters and visualize results for optimizing the algorithm of their choice using PDAL. Caveat- this notebook requires hard paths and the knowledge to build a Python environment and import packages. This notebook is also designed for SfM RGB point clouds. I will be adding another notebook for visualizing Micasense multispec point clouds.

### DIFF
--- a/contrib/iterate_DTM.ipynb
+++ b/contrib/iterate_DTM.ipynb
@@ -1,0 +1,328 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    " Joshua W. von Nonn \\\n",
+    " Geographer \\\n",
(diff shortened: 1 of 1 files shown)


pierotofy:
Hey @vonnonn :wave: thanks for the contribution! 

I wonder if instead of adding this to WebODM we should start a new "notebooks" repository for people to use.

Checking with @smathermather to see what he thinks.

vonnonn:
I like the idea of a separate notebook repo. Lemme know what you guys want to do.

pierotofy:
I've opened https://github.com/OpenDroneMap/Notebooks and given you write permissions @vonnonn (you can simply upload/commit directly to the repository there). I will close this PR since we're going to move notebooks contributions there 🙏

PR #1458, opened 2024-01-04
Update README.md

douw:
The instructions for running the the docker image as a Linux service is incorrect (possibly referring to running WebODM natively). These rewritten instructions refers to how I got WebODM Docker running on my RHEL 9 system


pierotofy (approved):
Nice, thanks @douw !

PR #1485, opened 2024-04-04
Ability to share particular map type

pierotofy:
You can now share links directly to elevation models, plant health, surface models, instead of the default orthophoto.


smathermather:
Oh heck yeah!

Issue #1065, opened 2021-10-02
Add imperial units to contours, measure tool


pierotofy:
We should add the ability to measure using imperial units, as well as export contours in values other than meters.

69chargr:
Would absolutely love this feature as I am based in the US. Pretty please? :-)

pierotofy:
Done!

PR #1496, opened 2024-05-01
Imperial units support, improvements, fixes, plugins API expansion

pierotofy:
Various improvements and fixes.


pierotofy:
Well, I started adding support for imperial units (both international and US), because it's overdue.

pierotofy:
#1065

smathermather:
[The Mars Climate Orbiter](https://en.wikipedia.org/wiki/Mars_Climate_Orbiter) will thank you.

pierotofy:
Also ~25% speed-up in rendering DSM/DTM tiles.

pierotofy:
This should be good to go.

smathermather:
Is it too late to ask for a couple more imperial units, specifically the following length and area measures?:

Area: gallons per fathom (GPF)

Length: hogsheads per rood (HPR)

ht @ivangayton

pierotofy:
Shouldn't be a problem, one can just modify `Units.js` and `UnitSelector.jsx`. 

https://www.youtube.com/watch?v=d9t5LqHcboI

smathermather:
> Shouldn't be a problem, one can just modify `Units.js` and `UnitSelector.jsx`. 

Super tempted to do that on crankyserver...

PR #1504, opened 2024-05-21
Minor improvements to Units support

pierotofy:
Work in progress.


pierotofy:
More sane layer list management (especially for project view, when there's many tasks):

image

Now supports grouping.

pierotofy:
Good to go.

PR #1506, opened 2024-05-27
Fix #1505 add a signal to notify that we are resizing images

### LINKED ISSUE #1505: Feature Request: Add a signal indicating that the images have been uploaded.
I am developing a plugin to obtain the align file from an API, but I don't have a way to get the signal from the system indicating that all images have been uploaded to create the Bounding Box of the images. I made some changes in the core to add that signal and retrieve it from the plugin. I would like to merge these changes into main.

diegoaces:


pierotofy:
Looks good to me, thanks @diegoaces !

PR #1514, opened 2024-06-20
remove line from hard recovery script

probably-not-porter:
the line `creating_task.console_output += gettext("Done!") + "\n"` in the process_task function (hard recovery guide) throws an error when following this process. console_output seems to be undefined in this context. 

Removing the line appears to result in a successful backup.


pierotofy:
Thanks!

Issue #921, opened 2020-10-04
Feature Request: View Current Location


Saijin-Naib:
### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Any modern HTML5 compliant browser (Firefox, Edge, Chrome, Safari)

### What is the problem?

I would like to see the ability for WebODM to take advantage of modern browser Geolocation APIs to allow for instant locating of the current user using a button on the GUI, with similar UX to established norms.

Example image as per Henrick:  
![image](https://user-images.githubusercontent.com/19295950/95027321-8ce1bd80-0665-11eb-9b2e-feaa828b7b22.png)

### What should be the expected behavior?

I believe this could be a candidate for a smaller funding drive.

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? Please include multiple screenshots of the problem! If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)

Thread:
https://community.opendronemap.org/t/current-position-webodm/5548

pierotofy:
Hi @Saijin-Naib,

It looks like there is some information missing from your issue that will be needed in order to diagnose and fix the problem at hand. Please take a look at the [Issue Template](https://github.com/OpenDroneMap/WebODM/blob/master/ISSUE_TEMPLATE.md), which will tell you **exactly** what your issue has to contain in order to be processable.

Also, double check that this is the right place. If you are just asking for information, reporting feedback or proposing a few feature, the right place to ask is the [Community Forum](http://community.opendronemap.org/c/webodm), not here.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next week (until 2020-10-07 16:15) I'll close this issue so it doesn't clutter the bug tracker.

Cheers!
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

chisquare88:
this function must need **https** protocol not http.

reference: [https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API)

pierotofy:
This is a cool idea for a plugin; we'd welcome pull requests! Anyone interested?

commanderwall:
Its a great idea! If it can work as a plugin.

commanderwall:
> This is a cool idea for a plugin; we'd welcome pull requests! Anyone interested?

[...]

PR #1515, opened 2024-06-25
Add GPS location plugin

pierotofy:
Adds a button for showing the current GPS location.

Requires WebODM to be run over https, so this is disabled by default.

image


pierotofy:
Closes #921

PR #1519, opened 2024-07-07
Update OpenStreetMap provider in Basemaps.js

danieldegroot2:
I would like to suggest using the now preferred https://tile.openstreetmap.org URL instead of the current one (`{s}.`), see

https://github.com/openstreetmap/operations/issues/737

Also, changes attribution URL to HTTPS link (use direct links). Furthermore, changes name to simply 'OpenStreetMap'. (use common name)

### DIFF
--- a/app/static/app/js/classes/Basemaps.js
+++ b/app/static/app/js/classes/Basemaps.js
@@ -20,10 +20,10 @@ export default [
   },
   {
     attribution:
-      '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>',
+      '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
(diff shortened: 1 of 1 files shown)


Saijin-Naib (commented):
Looks great, thanks! 

Have you tested locally?

danieldegroot2:
Fyi I'm on desktop browser only. I personally do (and can) not test such changes.
Based on existing use and documentation, it should work.
( see also https://github.com/openstreetmap/operations/issues/737#issuecomment-1318135343 , https://github.com/openstreetmap/operations/issues/737#issuecomment-1697013147 )
In one or two cases, HTTPS changes to URLs are discarded due to lacking support on supported browsers/devices.

Created also PR for regular ODM here: https://github.com/OpenDroneMap/ODM/pull/1775

PR #1523, opened 2024-07-09
fix: add blank=True to plugindatum's user field

samul-1:
In order to allow for global, plugin-wide settings, the PluginDatum's user field is nullable. However, the field doesn't come with the `blank=True` keyword argument. This makes it impossible to modify a global plugin setting from the admin panel, because the form will require the corresponding form field to contain a value.

This PR fixes this by adding `blank=True` to the `user` field, enabling administrators to change PluginDatum objects keeping the `user` value to null.


pierotofy:
Looks great, thanks @samul-1 !

Issue #1017, opened 2021-07-08
Feature Request: Image Geolocation Sanity Check


Saijin-Naib:
### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)
N/A

### What is the problem?
Sometimes, images in a dataset can have erroneous EXIF information that can cause reconstruction to fail in ways that seem mysterious and hard to track down (Child Returned 139, etc).
![Screenshot from 2021-07-08 18-14-21](https://user-images.githubusercontent.com/19295950/124999784-e7f17800-e01c-11eb-9791-480081a6feb0.png)

### What should be the expected behavior?
I think we should have an extra step after image selection/upload to WebODM that lists the images and their geolocation data (X,Y,Z, and rotation), as well as showing a mini-map. These two things should make it immediately obvious if any images in the set have bad geolocation data and should either be fixed manually or dropped from the reconstruction. I like the mini-map as pictured above in Geotagging, but also appreciate the no-nonsense tabular view of attributes from GPSCorrelate below.
![Screenshot from 2021-07-08 18-50-57](https://user-images.githubusercontent.com/19295950/125000044-782fbd00-e01d-11eb-9ac3-236455966189.png)

### How can we reproduce this? (What steps did you do to trigger the problem? What parameters are you using for processing? Please include multiple screenshots of the problem! If possible please include a copy of your dataset uploaded on Google Drive or Dropbox. Be detailed)
Use a dataset with erroneous data, like the one described above by jixue1.

J-xue:
Thank you a lot for Saijin-Naib's help on figuring out the issue, finding the root cause and propose this feature request.
Adding this feature before starting processing the images I think it's very important and helpful on finding the bias data at the first time, saving the total task performance time, improving the efficiency, success rate and accuracy. Integrate this feature to the WebODM instead of separating this step to other software, I think make the WebODM more robust and user friendly. Thank you.

PR #1530, opened 2024-07-23
Add flight path preview when selecting images

pierotofy:
Adds a plot of the flight path of the selected images (when GPS information is available). This is useful because it provides a sanity check on the selected images.

TODO:
 * Handle multiple image selects / drops
 * Add ability to specify a crop boundary directly from the map, instead of the tedious process of using geojson.io to specify a crop boundary via the task options.
 * (optional) I'm not sure I'm committed to do this right away, but could be neat to have the ability to remove images interactively, to remove mistakenly selected images.
 * Gracefully handle dataset with no georeferencing
 * Test very large datasets (e.g. more than 5000 images), exifr promises a ~2 seconds processing time, but I would test to see if it crashes...


pierotofy:
image

pierotofy:
Closes #1017

Saijin-Naib:
Love it! 

Is a tabular list also possible alongside the map?

pierotofy:
What would be the use case?

Saijin-Naib:
An easier (potentially) way to verify image geolocation and status, especially with mixed geoloc and non-geoloc data, without map view encompassing half the globe or worse to fit the preview extent. 

Also, thinking forward, might work for manual fixes/overrides to write out as geo.txt direct from this QA/QC step

pierotofy:
I could add buttons to export the data to GeoJSON / CSV. Could that work?

Saijin-Naib:
I think that covers some other excellent use-cases and simplifies other needed tools in the toolbox, so that is really cool for sure!

pierotofy:
Final implementation: 
image

Issue #1471, opened 2024-02-16
Update Django Framework from 2.2.27 to 4.2 LTS


chris-bateman:
### How did you install WebODM (docker, installer, etc.)?

N/A

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

N/A

### What is the problem?

Django framework is EOL.  WebODM uses 2.2.27 and the current LTS is 4.2
Placeholder issue to begin discussion on upgrading the framework.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

Review Django version here - https://github.com/OpenDroneMap/WebODM/blob/8150b822d9851fca434a5dedbdb5c5e1b3da2c3d/requirements.txt#L7

4.2 LTS makes sense support-wise, but v5 is also out.

chris-bateman:
EOL notes - https://www.djangoproject.com/download/ and https://endoflife.date/django

Saijin-Naib:
Do you know of any gotchas? API breaks? 

Keeping current on supported dependencies would be wonderful.

chris-bateman:
So this doco covers the process - https://docs.djangoproject.com/en/5.0/howto/upgrade-version/

Example incompatible notes - https://docs.djangoproject.com/en/5.0/releases/4.2/#backwards-incompatible-changes-in-4-2

And a good list here as well - https://docs.djangoproject.com/en/5.0/internals/deprecation/

chris-bateman:
I started to make a little progress on this.  I am playing dependency wack a mole.
So with that in mind it might be time to audit the current modules located in this file - 
https://github.com/OpenDroneMap/WebODM/blob/master/requirements.txt

The main focus is identifying packages that are deprecated/abandoned with no clear support from their maintainer/community.
Some have suitable alternatives noted in their doco but others will require some research to find a suitable replacement.

**django-guardian**
![image](https://github.com/OpenDroneMap/WebODM/assets/12206103/6e8d202d-9f99-4e13-a475-52edd01e7b9b)

**rfc3987** - Github repo was archived on the 12th Jan 2024 but moved to a new repo recently.
https://codeberg.org/atufi/rfc3987/
![image](https://github.com/OpenDroneMap/WebODM/assets/12206103/64541de4-0683-454e-bb73-470c7e5d1f0d)

[...]

Issue #826, opened 2020-02-27
Upgrade to PostgreSQL 12.x


Saijin-Naib:
Similar to the move to Python3 and possibly to latest LTS (20.04), PostgreSQL 9.x is deprecated above 16.04 LTS and is replaced by PostgreSQL 10.x on 18.04.3 and 12.x on 20.04.

18.04.3:  
https://packages.ubuntu.com/bionic/database/postgresql

20.04:  
https://packages.ubuntu.com/focal/database/postgresql

pierotofy:
Hi @Saijin-Naib,

It looks like there is some information missing from your issue that will be needed in order to diagnose and fix the problem at hand. Please take a look at the [Issue Template](https://github.com/OpenDroneMap/WebODM/blob/master/ISSUE_TEMPLATE.md), which will tell you **exactly** what your issue has to contain in order to be processable.

Also, double check that this is the right place. If you are just asking for information, reporting feedback or proposing a few feature, the right place to ask is the [Community Forum](http://community.opendronemap.org/c/webodm), not here.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next week (until 2020-03-01 15:00) I'll close this issue so it doesn't clutter the bug tracker.

Cheers!
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

Saijin-Naib:
Thanks again, Piero. I still don't understand what dependencies are for each component only 🤣

Of course, we might be able to avoid the whole PGSQL upgrade dance with OGC GeoPackage ;)
https://community.opendronemap.org/t/ogc-geopackage-for-data-storage-back-end/3650?u=saijin_naib

ssantacroce-uwach:
@pierotofy I figured I would take a stab at this, since 9 is SUPER old... any reason we are using our own dockerfile vs using a base postgis image like postgis/postgis:14-3.3 and throwing in scripts in the init.db folders?? 

Switching to a base image should make it a lot easier to upgrade versions

pierotofy:
We use our own (compiled from source) version of postgres/postgis, because they don't publish 9.x releases on docker hub anymore.

The problem with postgres/postgis, using docker, is that there's no easy way to perform a fully automated migration for dozens of thousands of users that are using WebODM via docker.

The correct way to perform the migration, is to create a new docker image which has the newer postgres/postgis version and perform the migration on startup. Postgres does not offer an automated way to do this, so it's going to be a bit tricky.

[...]

PR #1543, opened 2024-08-19
refactor: Update Django to 4.2 LTS

NtskwK:
# What's new

## Docker-compose:

[Version top-level element (obsolete)](https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md)

## Django 4.2:

[Django 4.2 supports PostgreSQL 12 and highe](https://docs.djangoproject.com/en/5.0/releases/4.2/#dropped-support-for-postgresql-11)

[django.conf.urls.url() is removed.](https://docs.djangoproject.com/en/4.1/releases/4.0/#features-removed-in-4-0)

[django.utils.encoding.force_text() and are removed.smart_text()](https://docs.djangoproject.com/en/4.2/releases/4.0/#:~:text=django.utils.encoding.force_text()%20and%20are%20removed.smart_text())

[PostgreSQL JSONField is deprecated](https://docs.djangoproject.com/en/5.1/releases/3.1/#postgresql-jsonfield)

## Others

PostGIS Support Matrix: https://trac.osgeo.org/postgis/wiki/UsersWikiPostgreSQLPostGIS

### djangorestframework-simplejwt
there are also many different from [django-rest-framework-jwt](https://github.com/jpadilla/django-rest-framework-jwt) to [djangorestframework-simplejwt](https://github.com/jazzband/djangorestframework-simplejwt). So it can not pass the test in webodm.sh.

### rio-tiler

RioTilerBaseModel:warn 'key' access will has been deprecated and will be removed in rio-tiler 7.0."

### webpack

- (node:5096) [DEP_WEBPACK_DEPRECATION_ARRAY_TO_SET] DeprecationWarning: Compilation.chunks was changed from Array to Set (using Array method 
'map' is deprecated)

- (node:5096) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

- WARNING in webpack performance recommendations: webapp | You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.

### django-webpack-loader
There is a bug(?) in static template at v1.0.0 or higher version.

```
Error during template rendering In template /mnt/d/Program/gitwork/WebODM/app/templates/app/base.html
```

### Docker-compose

[Maintainer tag is deprecated](https://docs.docker.com/reference/build-checks/maintainer-deprecated/)


NtskwK:
#1471

pierotofy:
Thanks @NtskwK :wave: I haven't had a chance to review everything, but a few things stand out right away:

 * Please do not remove the database migrations! This would break updates from older versions.
 * It's nice to update Postgres to a newer version (ref #826), but I do not see a mechanism for performing a seamless update when users issue `./webodm.sh update`. This is a requirement for considering updating the database, and we are not going to break the installation for thousands of people. See this comment here https://github.com/OpenDroneMap/WebODM/issues/826#issuecomment-1275737159

NtskwK:
Yes, an automated scripts souds good.
I see those files are not compatible here. It must be solved.
```
db\Dockerfile
...
> 83 | VOLUME /var/lib/postgresql/data
```
I have never tried to migrate database with data. Could you give me some suggest？

NtskwK:
I found some manual about it. But I have no idea for it.

https://www.postgresql.org/docs/current/pgupgrade.html

PR #1544, opened 2024-08-20
group  installations instructions 

sbonaime:


pierotofy:
Looks good, thanks @sbonaime !

PR #1545, opened 2024-08-25
refactor: update Django to 4.2 LTS

NtskwK:
# What's new

## Django 4.2:

[Django 4.2 supports PostgreSQL 12 and highe](https://docs.djangoproject.com/en/5.0/releases/4.2/#dropped-support-for-postgresql-11)

[django.conf.urls.url() is removed.](https://docs.djangoproject.com/en/4.1/releases/4.0/#features-removed-in-4-0)

[django.utils.encoding.force_text() and are removed.smart_text()](https://docs.djangoproject.com/en/4.2/releases/4.0/#:~:text=django.utils.encoding.force_text()%20and%20are%20removed.smart_text())

[PostgreSQL JSONField is deprecated](https://docs.djangoproject.com/en/5.1/releases/3.1/#postgresql-jsonfield)

[Activate the postgis_raster extension with Django. ](https://docs.djangoproject.com/en/5.1/ref/contrib/gis/install/postgis/#creating-a-spatial-database)

## Docker-compose:

[remove Version top-level element (obsolete)](https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md)

## Docker

set `ENV` with `=` (e.g. *ENV VAR=xxx*)

## Others

PostGIS Support Matrix: https://trac.osgeo.org/postgis/wiki/UsersWikiPostgreSQLPostGIS

### djangorestframework-simplejwt **(Critical)**
there are also many different from [django-rest-framework-jwt](https://github.com/jpadilla/django-rest-framework-jwt) to [djangorestframework-simplejwt](https://github.com/jazzband/djangorestframework-simplejwt). ~~So it can not pass the test in `webodm.sh`.~~ Login function is broken now. The browser will create `window.prompt` automatic which is required account and password to instead of the login function. It's necessary to overwrite auth functions.

### django-webpack-loader
There is a bug(?) in static template at v1.0.0 or higher version.

```
Error during template rendering In template /mnt/d/Program/gitwork/WebODM/app/templates/app/base.html
```

### rio-tiler

RioTilerBaseModel:warn 'key' access will has been deprecated and will be removed in rio-tiler 7.0."

### webpack

- (node:5096) [DEP_WEBPACK_DEPRECATION_ARRAY_TO_SET] DeprecationWarning: Compilation.chunks was changed from Array to Set (using Array method 
'map' is deprecated)

- (node:5096) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

- WARNING in webpack performance recommendations: webapp | You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.

### Docker-compose

[Maintainer tag is deprecated](https://docs.docker.com/reference/build-checks/maintainer-deprecated/)


NtskwK:
@pierotofy 
#1471 #1543

Dockerfile
```diff
@@ -1,33 +1,55 @@
-FROM ubuntu:21.04
+FROM ubuntu:20.04
```
chris-bateman, line 1: Curious about the move from 21 to 20 here.  Due to the upcoming EOL of 20.04 would it make sense to work on bumping this to 22.04 or higher?

NtskwK: Version of ubuntu in Docker `webodm_db` and `webodm_webapp` should be same to ensure apt will install some package in the same version. Maybe I should to update both?

NtskwK: Done. The base image is Ubuntu 22.04 now, because PDAL can not be found by apt in Ubuntu 24.04.

chris-bateman: Yeah 24 is still fairly new but 22 has plenty of life left in it.

pierotofy:
I appreciate the efforts to upgrade Django, Postgres and base system, but there's a lot that needs to be addressed.

My comments from https://github.com/OpenDroneMap/WebODM/pull/1543#issuecomment-2297562263 still haven't been addressed and I don't understand why this was marked as "ready for review".

This is not an easy task and there's a reason it hasn't been done yet; it's a lot of work and needs a solid amount of funding (or a lot of careful work by a dedicated contributor) to be done.

PR #1546, opened 2024-08-29
Update README.md solve issue on first start on windows

sbonaime:
solve issue on first start on windows


pierotofy:
Should this be added to the common troubleshooting table? 

https://github.com/OpenDroneMap/WebODM?tab=readme-ov-file#common-troubleshooting

sbonaime:
I don't know if it is a known issue or the way to do it

Issue #1237, opened 2022-09-10
Installing WebODM in a different folder causes the node to be offline


pierotofy:
This is due to how docker assigns hostnames. The line in webodm.sh:

```
NODE_HOST=$(python manage.py getnodehostname webodm_node-odm_$i)
```

Should probably reflect the directory name.

pierotofy:
This *should* be fixed.

PR #1550, opened 2024-09-03
feat: try to fix default hostname

NtskwK:


pierotofy:
Hey thanks @NtskwK, I understand what the intent here is, but can you provide some context about how to reproduce the problem this patch is solving?

NtskwK:
I noticed that the name of container are not always the same when I try to deploy the WebODM on WSL with `./webodm.sh start --dev --gpu(?)` . Sometime, `node-odm-1`, `webodm-node-odm-1`, `[DIR_NAME]-odm-1`.

I don't fully understand the way nodeodm is named. But setting the hostname to the container name will allow them to connect successfully.

pierotofy:
I wonder if explicitly setting the `container_name` (https://docker-docs.uclv.cu/compose/compose-file/compose-file-v2/#container_name) property to `docker-compose.nodeodm.yml` would make the hostname predictable and not require runtime checks.

NtskwK:
That's weird. If there is already a container named `webodm_db`/`webodm_webapp`/`broker`/`worker`, docker-compose will throw an error `xxx is already exist` then stop batch. But it's different to `node-odm-1`, because `docker-compose.nodeodm.yml`(and `.gpu.*`) dose not appoint container_name with `container_name: xxx`. So [DIR_NAME]-node-odm-1 will be named by Docker automaticlly.

NtskwK:
```
> webodm.sh

down(){
	command="$docker_compose -f docker-compose.yml"

	if [ "${GPU_NVIDIA}" = true ]; then
		command+=" -f docker-compose.nodeodm.gpu.nvidia.yml"
	elif [ "${GPU_INTEL}" = true ]; then
		command+=" -f docker-compose.nodeodm.gpu.intel.yml"
	else
		command+=" -f docker-compose.nodeodm.yml"
	fi

	command+=" -f docker-compose.nodemicmac.yml down --remove-orphans"

	run "${command}"
}
```

`docker-compose.nodeodm.gpu.*` will not be shutdown if `--gpu` was not be used in `./webodm down`. This can be an important cause of container residue. Then, a new WebODM program starter will launch a container which is named  `DIR_NAME]-node-odm-1`.

NtskwK:
[On user-defined networks like alpine-net, containers can not only communicate by IP address, but can also resolve a container name to an IP address.](https://docs.docker.com/engine/network/tutorials/standalone/#:~:text=resolve%20a%20container%20name%20to%20an%20IP%20address)

[Is can only works in `user-defined bridge networks` and disable in `default bridge networks`.](https://docs.docker.com/engine/network/tutorials/standalone/#use-the-default-bridge-network:~:text=This%20succeeds.%20Next,This%20will%20fail.)

NtskwK:
> 我想知道将 （https://docker-docs.uclv.cu/compose/compose-file/compose-file-v2/#container_name） 属性设置为是否会使主机名可预测并且不需要运行时检查。`container_name``docker-compose.nodeodm.yml`
> 
> 如果是这样，我们可以在 is in 时使用它（但不能在 is > 1 时使用它）。`--default-nodes``1``webodm.sh``--default-nodes`

On user-defined networks like alpine-net, [containers can not only communicate by IP address, but can also resolve a container name to an IP address.](https://docs.docker.com/engine/network/tutorials/standalone/#:~:text=resolve%20a%20container%20name%20to%20an%20IP%20address)

Is can only works in `user-defined bridge networks` and [disable in `default bridge networks`.](https://docs.docker.com/engine/network/tutorials/standalone/#use-the-default-bridge-network:~:text=This%20succeeds.%20Next,This%20will%20fail.)

NtskwK:
> I wonder if explicitly setting the `container_name` (https://docker-docs.uclv.cu/compose/compose-file/compose-file-v2/#container_name) property to `docker-compose.nodeodm.yml` would make the hostname predictable and not require runtime checks.
> 
> If so, we could use it when  `--default-nodes` is `1` in `webodm.sh` (but not when `--default-nodes` is > 1).

It's theoretically possible, but not recommended. The logic of the function will become complex. If docker-compose starts successfully, but docker-compose.node is blocked due to duplicate names, it will be deployed in a state without nodes, which is not expected.

pierotofy:
https://docs.docker.com/compose/networking/

Mm, perhaps just setting COMPOSE_PROJECT_NAME to "webodm" would do?

NtskwK:
Of course that's okay.

Please turn it into a draft. I will try to modify the webodm.sh script in passing.
________________________________
发件人: Piero Toffanin ***@***.***>
发送时间: 2024年9月5日 0:48
收件人: OpenDroneMap/WebODM ***@***.***>
抄送: NtskwK ***@***.***>; Mention ***@***.***>
主题: Re: [OpenDroneMap/WebODM] feat: try to fix default hostname (PR #1550)


https://docs.docker.com/compose/networking/

Mm, perhaps just setting COMPOSE_PROJECT_NAME to "webodm" would do?

―

NtskwK:
#1237

webodm.sh
```diff
@@ -576,13 +572,11 @@ elif [[ $1 = "stop" ]]; then
...
-		command+=" -f docker-compose.nodeodm.gpu.nvidia.yml"
-	elif [ "${GPU_INTEL}" = true ]; then
-		command+=" -f docker-compose.nodeodm.gpu.intel.yml"
-	else
-		command+=" -f docker-compose.nodeodm.yml"
-	fi
+
+	command+=" -f docker-compose.nodeodm.gpu.nvidia.yml"
+	command+=" -f docker-compose.nodeodm.gpu.intel.yml"
+	command+=" -f docker-compose.nodeodm.yml"
```
pierotofy, line 578: I'm not sure I understand the change here.

docker-compose.nodeodm.gpu.nvidia.yml
```diff
@@ -11,6 +11,7 @@ services:
       - WO_DEFAULT_NODES
   node-odm:
     image: opendronemap/nodeodm:gpu
+    container_name: node-odm-1
```
pierotofy, line 14: Did you test with:

`./webodm.sh start --default-nodes 3` ?

webodm.sh
NtskwK: The shutdown function now has to be entered after the `--gpu` to shut down the gpu node. But it's just the startup parameter in the manual. There may be users who get unexpected results. I would like to be able to make sure they shut down.

docker-compose.nodeodm.gpu.nvidia.yml
NtskwK: My apologies. I forgot there would be conflict here. 

I still don't fully understand the way to the naming. Can you give me some advice?

pierotofy: You need to understand how docker-compose works.

This is a good start: https://docs.docker.com/compose/

Read it all.

NtskwK:
You are right, maybe I should not affect the behavior of the other. Just focus on the try to fix default hostname like next function do.
> Workaround for mysterious "webodm_node-odm-1" or "webodm-node-odm-1" hostname switcharoo on Mac

How do you think？

Issue #1528, opened 2024-07-21
Duplicating task prior to processing results in tasks that are entangled


smathermather:
### How did you install WebODM (docker, installer, etc.)?

Docker deploy on Ubuntu 24.04

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Brave, Chromium, Firefox

### What is the problem?

If one uploads images to a new task, processes the data, and the duplicates the task, everything works as expected.

If one uploads images to a new task, cancels the task before finishing processing, and then duplicates the canceled task, WebODM seems to confuse the two tasks, for example, restarting each task will result in seeing the same logging showing in each task.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

If one uploads images to a new task, cancels the task before finishing processing, and then duplicates the canceled task, WebODM seems to confuse the two tasks, for example, restarting each task will result in seeing the same logging showing in each task.

Happens with any dataset, but to keep the bot happy, here's a link:
link.com

lfeldb:
I can confirm this issue.
Also it does not only seem to happen on canceling the task manually, but also when tasks are completed (succesfully with an error) and then copied.

I also just noticed when i create a copy, the field 'name' is not updated in /var/www/data/tasks.json in nodeodm docker container. It seems that this file is created not at start-time, but later on in the process.

What does the 'off-by-one error' tag mean? Can I assist with some way?

lfeldb:
OK, I'll look into it as soon as I have some time. Thank you for the pull request reference, this saves me some time getting familiar with the code.

( Alright! I thought off-by-one might have suggested an indexing error, or it was one fix away from being fixed :-)

smathermather:
> OK, I'll look into it as soon as I have some time. Thank you for the pull request reference, this saves me some time getting familiar with the code.

Awesome. Thanks!

> ( Alright! I thought off-by-one might have suggested an indexing error, or it was one fix away from being fixed :-)

It was a tag I added a long time ago as a joke, and sometimes it is at least almost apropos. :slightly_smiling_face:

pierotofy:
Should be fixed with #1559

Issue #1416, opened 2023-10-09
SQL exception when adding a user with a quota


argoopen:
### How did you install WebODM (docker, installer, etc.)?

Docker

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Client: Firefox latest (118.0.1); MacOS Ventura
Server: Alma Linux 9

### What is the problem?

I've tried installing twice on two separate servers via Docker, and under both instances whenever I add a new user account with a quote other than -1 I see an error: 500 Internal Server Error. Something happened. The server logs contain more information.

I see no issues when adding an account without a quota (quota = -1).

Updating a user's quote after adding the account works as expected.

In my logs I see a Postgres SQL exception:

ERROR:  duplicate key value violates unique constraint "app_profile_user_id_key" 
db                 | DETAIL:  Key (user_id)=(9) already exists. 
db                 | STATEMENT:  INSERT INTO "app_profile" ("user_id", "quota") VALUES (9, 500.0) RETURNING "app_profile"."id"

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

 - click 'Administration'
 - click 'Accounts'
 - click 'Add user'
 - enter username and password
 - enter any positive value for Quote (eg 500, or 1000)
 - click any of the Save buttons (Save and add another, Save and continue editing, or SAVE)

pierotofy:
Thanks @argoopen :pray: this should be mitigated as part of #1559, we now hide the "quota" field during user creation (it can be edited after creating the user). The django admin's functionality doesn't seem able to handle editing of fields during user creation that have been setup the way we have (with a User <=> Profile 1:1 relationship).

PR #1559, opened 2024-09-28
Create BIGTIFF output during export if needed

pierotofy:
Currently larger TIFF outputs will fail to export with a `Chunk and warp failed` error message when the file exceeds 4GB.


pierotofy:
Also fixes #1528

pierotofy:
Closes #1416

PR #1568, opened 2024-11-06
Public edits

pierotofy:
Adds the ability to allow shared links to be edited, e.g. allows Potree annotations to be made by external users just with a link.

This has been asked a few times before.


smathermather:
Very nice.

PR #1570, opened 2024-11-20
Fix: Silence Deprecated Version Warning

Saijin-Naib:
Remove deprecated Docker Version object from YML files


pierotofy:
But will this break with older versions of docker-compose?

Saijin-Naib:
I need to package it to test, Alpine doesn't ship v1 since it is long since deprecated.

One of the major improvements of v2 is using ```-``` instead of ```_``` for the containers to ensure the hostnames are DNS-compliant and reachable, which seems like a very important UX improvement vs keeping the older compose.

kikislater:
+1 for this PR
v1 is out of date now, it is time to update !

PR #1578, opened 2025-01-06
Handle import chunk retries

pierotofy:
Adds the ability to retry chunk import uploads, which is important in environments where network connectivity is intermittent.

### DIFF
--- a/app/api/tasks.py
+++ b/app/api/tasks.py
@@ -471,6 +471,7 @@ class TaskAssetsImport(APIView):
                 os.unlink(tmp_upload_file)
             
             with open(tmp_upload_file, 'ab') as fd:
+                fd.truncate(byte_offset)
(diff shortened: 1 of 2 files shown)


Saijin-Naib:
This is HUGE, but in little chunks 🤣

pierotofy:
Yeah I'm actually surprised I didn't do this sooner, Dropzone already had a function to retry chunks, but I just never enabled it (didn't know it was there).

Saijin-Naib:
This is a massive stack 🤷 One person can't know each lib top to bottom and left to right. 

I'm excited to see how this helps folks out. 

Does this warrant a minor version bump in WebODM?

pierotofy:
> Does this warrant a minor version bump in WebODM?

Nah, all part of 2.5.7.

Issue #1582, opened 2025-01-10
Imported task from backup lacks "Restart" button


pierotofy:
An imported backup lacks the "Restart" button for a task (because of a `imported = task.import_url !== "";` check in TaskListItem.jsx), but really should have it.

Issue #1581, opened 2025-01-10
3D button in maps popup leads to 404 page


pierotofy:
Steps:

1. Share a task's map
2. Open the map via the shared link
3. Click on the orthophoto
4. Popup opens
5. Click the 3D button from the popup
6. Witness: 404

Issue #1580, opened 2025-01-10
Exporting large DEMs rasters with Shading takes a long time and memory


pierotofy:
Even after the improvements of #1579, the software still struggles to export colored DEMs with hillshading in a reasonable time and under reasonable memory bounds.

PR #1583, opened 2025-01-10
More memory optimizations for RGB/HSV conversion, duplicate filename uploads

pierotofy:
More optimizations that almost 2x improve performance in RGB/HSV conversion and substantially reduce memory usage.

Could mitigate #1580


pierotofy:
Closes #1581

pierotofy:
Closes #1582

PR #1584, opened 2025-01-11
More consistent hillshading across zoom levels

pierotofy:
I realized that the dynamic tiler applies hillshading inconsistently, e.g. at higher zoom levels the shading appears "lighter" than lower zoom levels. This can be noticed when zooming in/out. This is because the formula for `delta_scale` was linear, instead of quadratic, which is incorrect.

This would also lead to inconsistencies when exporting hillshaded DEMs, where the hillshading looked particularly "light" instead of looking like on-screen.

### DIFF
--- a/app/api/hillshade.py
+++ b/app/api/hillshade.py
@@ -71,12 +71,6 @@ class LightSource:
             A 2d array of illumination values between 0-1, where 0 is
             completely in shadow and 1 is completely illuminated.
         """
-
-        # Because most image and raster GIS data has the first row in the array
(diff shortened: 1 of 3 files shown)


pierotofy:
master

pierotofy:
There's also a consistency problem with color scaling, since during export we have white background that pushes the max value of the RGB data to 255, whereas most colormaps don't have pure white. Working on a fix for that..

pierotofy:
Current map view (left) vs export (right):

image

After disabling contrast stretching:

image

PR #1585, opened 2025-01-13
JS bundle optimization, placeholder loading screens

pierotofy:
A few optimizations to reduce the size of the js bundles.


pierotofy:
Aside from several graphical fixes, the coolest addition of this PR are placeholder loading screens:

image

This will make the feel of the UI much snappier as WebODM loads code and models.

Issue #1586, opened 2025-01-16
WebODM won't start - repeated webapp exception


rahlhar:
### How did you install WebODM (docker, installer, etc.)?

Installed `docker v27.3.1`. Used `git clone` followed by `./webodm.sh start`.

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

Browser is irrelevant - the problem is a failure to start.

OS is `Artix Linux` running `6.12.9-hardened` (confirmed also on `6.12.9`).

### What is the problem?

When running `./webodm.sh start`, all seems to flow well until the `webapp` process keeps crashing, retrying, ad nauseum.

```
webapp      | celery beat v4.4.0 (cliffs) is starting.                                                                                         
webapp      | [2025-01-16 10:16:05 +0000] [240] [INFO] Starting gunicorn 19.7.1                                                                
webapp      | Traceback (most recent call last):                                                                                               
webapp      |   File "/usr/local/lib/python3.9/dist-packages/gunicorn/sock.py", line 44, in set_options                                        
webapp      |     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)                                                                   
webapp      | OSError: [Errno 95] Operation not supported                                                                                      
webapp      |                                                                                                                                  
webapp      | During handling of the above exception, another exception occurred:                                                              
webapp      |                                                                                                                                  
webapp      | Traceback (most recent call last):                                                                                               
webapp      |   File "/usr/local/bin/gunicorn", line 8, in <module>                                                                            
webapp      |     sys.exit(run())                                      
webapp      |   File "/usr/local/lib/python3.9/dist-packages/gunicorn/app/wsgiapp.py", line 74, in run                                         
webapp      |     WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()                                                                     
webapp      |   File "/usr/local/lib/python3.9/dist-packages/gunicorn/app/base.py", line 203, in run                                           
webapp      |     super(Application, self).run()                                                                                               
webapp      |   File "/usr/local/lib/python3.9/dist-packages/gunicorn/app/base.py", line 72, in run                                            
webapp      |     Arbiter(self).run()                                                                                                          
webapp      |   File "/usr/local/lib/python3.9/dist-packages/gunicorn/arbiter.py", line 198, in run
webapp      |     self.start()                                         
[...]

PR #1587, opened 2025-01-19
Update gunicorn to 19.8.0 fixing #1586

hauke96:
As the title says, this fixes issue #1586. I tested this change by building a docker image (using `docker-compose.build.yml`) containing the change and then basically starting the image with the `webodm.sh` script. The container actually contained the correct `sock.py` file (which contains the problematic code causing the issue in the old version) and starts without problems. I also tried the web-interface and used WebODM on a set of images. Everything worked great.


pierotofy:
Thanks for the PR @hauke96 !

Issue #1567, opened 2024-10-28
Cesium Ion Plugin activated prevents loading the task contents of the dashboard in any project


luismgarciac:
### How did you install WebODM (docker, installer, etc.)?

Windows 11 Home, WSL, Ubuntu 22.04.5 LTS, Docker Desktop
WebODM 2.5.5

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)

[Edge 130 on Windows 11](https://www.whatismybrowser.com/detect/what-version-of-edge-do-i-have/)
[Chrome 130 on Windows 11](https://www.whatismybrowser.com/detect/what-version-of-chrome-do-i-have/)

### What is the problem?

I am trying to use the Cesium Ion Plugin to automatically upload and georeference a 3D textured model created with WebODM, but when I activate the Cesium and go to the dashboard I am unable to open any dataset from any model, the menu of the dataset goes blank, so I am not even able to try the Cesium Ion Plugin.

However, if I upload the data directly to Cesium Ion, the data is visible but not georeferenced, and has to be georeferenced manually.

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)

I have uploaded to Google Cloud a dataset that I intend to use for this purpose, but any other dataset produces the same results:
https://drive.google.com/file/d/1TYiYmjVLzHNGAJT1A5i73op-Ft1ZPVQQ/view?usp=drive_link

Before processing I made sure that the texturing-single-material setting was enabled as indicated in the Cesium Ion integration documentation:
https://cesium.com/learn/ion/integrations-learn/integrations-webodm/

When the Cesium Ion plugin is enabled the dashboard content goes blank as one project task is expanded:
![image](https://github.com/user-attachments/assets/53054d30-57c0-4df7-b7a0-726332d50cba)

When the Cesium Ion plugin is disabled the dashboard content becomes visible, but it is not possible to use the plugin:
![image](https://github.com/user-attachments/assets/ad5f44a0-024d-4eef-b777-ead17d729159)

The developer console provides the following error message:
![image](https://github.com/user-attachments/assets/c52adc42-e698-4f22-b212-023197278a9e)
`main-513c83c….js:2 
 TypeError: Cannot read properties of undefined (reading 'POINTCLOUD')
    at eval (TaskView.js:2:16907)
[...]

PR #1597, opened 2025-02-06
Simultaneous map items display and swipe view

### LINKED ISSUE #781: Allow orthophoto, plant health and surface models to be displayed simultaneously + swipe view
Good improvements would include:

 * Allow a user to add orthophoto, plant health and DEMs at the same time
 * Add swipe view to compare layers, UAVArena style

pierotofy:
Closes #781


pierotofy:
*Might* fix #1567, at the very least it doesn't crash anymore.

Issue #1612, opened 2025-02-19
New configurable login page


diegoaces:
I would like to send this change that could be useful.
@pierotofy 

![Image](https://github.com/user-attachments/assets/68fc5a55-58b2-43e1-82d2-3b679fbb2933)

https://community.opendronemap.org/t/proposal-new-configurable-login-page/23453

PR #1613, opened 2025-02-19
Custom login page

diegoaces:
#1612


pierotofy:
Thanks @diegoaces ! This will look nice indeed.

A few requests before merging:

 - [ ] Please merge the migrations into a single file
 - [ ] Don't duplicate `base.html` to create `base_login.html`, I understand you probably did this in order to not display the top menu, but I'm not sure we should remove the menu entirely (at least give users the option to keep/display the menu in the login screen)?
 - [ ] There are multiple DOM elements with ID slogan-text; IDs should be unique. Use classes instead 
 - [ ] If a slogan is not set, the corresponding h3 element should not be displayed.
 - [ ] Rename #column-left --> #login-column-left
 - [ ] Can we avoid creating a `registration_base_login.html`? I don't understand why these changes require so many new templates.

Also, does this display nicely on mobile / portrait mode?

diegoaces:
Thank you, @pierotofy, for your feedback.
Based on your comments, I have made the following changes:

- Removed base_login.html
- Removed slogan-text IDs.
- Removed registration_base_login.html
- If a slogan is not set, the corresponding h3 element should not be displayed.
- Added login-column-left and login-column-right to css
- Support to mobile/portrait mode.
- Removed icon size setting.

 If I missed anything, please let me know.

pierotofy:
Thanks for the changes @diegoaces !

I started doing some review and found a potentially problematic issue with the proposed new layout; sorry I didn't think of this sooner. The proposed layout assumes that people have uploaded a sufficiently high resolution image for a logo, but this is not necessarily the case, as shown below for one example of a low resolution image:

image

Mobile rendering:

image

Leaving the default logo size definitely looks a bit odd too:

image

I most definitely do not want to force thousands of users to upload new high resolution images to make their login screen look good.

Another related side-issue: if the header background is white, the whole page looks a bit odd also:

image


Perhaps the solution here could be to offer the option to switch between the two login layouts (legacy vs column), leaving existing users on the old layout by default, but I'm not sure the additional complexity would be worth the effort?

I do start to wonder if perhaps this change falls under custom code that perhaps should be best kept in a fork; designing generic layouts that can be customized can be tricky...

pierotofy:
Closing for now.

PR #1623, opened 2025-03-08
Add hosted instance link alert

pierotofy:
image

This comes up so often in forum discussions, hopefully this will help guide people.


kikislater:
10.x.x.x is also private and common !

jet86:
If we want to be thorough, we should also detect IPs in the 172.16 /12 range (ie 172.16.0.0 - 172.31.255.255) as then we'd be detecting all private addresses as per rfc1918. Unfortunately that's not quite as simple as the 10 /8 and 192.168 /16 ranges. I could submit a PR that individually checks each of the 16 x /16 subnets in the 172.16 /12 range, but I'm hoping someone else knows a more elegant way of checking that range in a single line?

PR #1626, opened 2025-03-11
Add 172.16 /12 as Private IP Range

jet86:
Adds a check for the 172.16 /12 private address range so all IPv4 private IP ranges are now checked.

Also re-orders the checks so that they are grouped a bit more logically (IPv4 default route, then localhost addresses, then private IP ranges, then IPv6 localhost).

cf [PR#1623](https://github.com/OpenDroneMap/WebODM/pull/1623)


pierotofy:
Thanks @jet86 !

PR #1634, opened 2025-03-23
Rename node-odm Docker container if needed with rename_containers.sh

livmion:
Dear maintainers,

as discussed in the [WebODM forum](https://community.opendronemap.org/t/no-connection-to-local-webodm-node-1-after-clean-installation/23752/8), I noticed that sometimes the Docker containers for the nodes do not connect correctly to the application. This behavior is caused by Docker assigning the names with hyphens (`-`) instead of underscores (`_`). The bug affects Debian 13 (Trixie) with `docker-compose 2.26.1`.

In order to avoid this problem, I edited the `webodm.sh` file and added a new script, called `rename_containers.sh`. The changes are the following:

- Instead of creating and running the containers at the same time, they are just created by using the `--no-start` option
- The script rename_containers.sh is initialized and checks if containers with the prefix `webodm-node-odm-` are present. If so, it renames them to `webodm_node-odm_`
- Afterwards, it runs the containers as before with the correct DNS

All the best

Alessio

### DIFF
--- a/rename_containers.sh
+++ b/rename_containers.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Short script to correct Docker inconsistent behavior with `node-odm` names.
+# WebODM expects DNS names in the Docker network for `node-odm` to be something
+# like `webodm_node-odm_`. Docker standard names use hyphens `-` instead of
+# underscores `_` and DNS name are created accordingly. This discrepancy
+# causes the connection failure to the `node-odm` containers.
+# This script checks if the container names follow the unexpected form
+# `webodm-node-odm-` and change them to `webodm_node-odm_`. It should work
+# correctly even if many different `node-odm` instances are started.
+
+dc_msg_ok="\033[92m\033[1m OK\033[0m\033[39m"
+
+BASE_NAME="webodm-node-odm-"
+NEW_BASE_NAME="webodm_node-odm_"
+
+# Create a list with all containers matching the pattern for wrong names.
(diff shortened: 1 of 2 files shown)


pierotofy:
Hey thanks for the PR @livmion 🙏 

I'm familiar with this problem and we actually have in place a mechanism for renaming the node's hostname just like you described/implemented; the real question is, why didn't the built-in mechanism fix the hostname correctly in your case?

https://github.com/OpenDroneMap/WebODM/blob/master/worker/tasks.py#L47

pierotofy:
In reality the *real* fix for this would be to just accept that docker has unreliable hostnames across versions, and the only reliable thing to do would be to manually specify the container_name and deprecate/remove the ability to setup multiple nodes via `--default-nodes`. It makes little sense to scale on the same machine anyway. We should expose a method for increasing the NodeODM queue size rather than setting up multiple NodeODM containers on the same machine.

pierotofy:
In order to debug this further, you would need to figure out:
 * Why is WebODM picking `webodm-node-odm-` as the hostname, even though we always first try with `webodm_node-odm_`. See https://github.com/OpenDroneMap/WebODM/blob/master/start.sh#L71 and https://github.com/OpenDroneMap/WebODM/blob/master/app/management/commands/getnodehostname.py
 * Is https://github.com/OpenDroneMap/WebODM/blob/master/worker/tasks.py#L47 getting triggered at all / is it the cause of the problem?
 * Is it a race condition (is the node-odm container running when the codes above execute?)

pierotofy:
I'll close this for now, but feel free to re-open a new PR once more is known. 🙏

PR #1638, opened 2025-03-30
Add IPV6 Support

as598466:
Hello everyone,

In an effort to reduce costs (IPv4 shortage) and to align with a more modern network architecture, I would like to be able to use IPv6 with the WebODM/NodeODM pair.

Currently, I host a WebODM instance in Docker on a Debian 12 Linux server. When I need to generate a project, I create a remote NodeODM server and add the node to my WebODM server. All of this works fine in IPv4 (by default). However, when I add a NodeODM node that is only accessible via IPv6 (taking care to add the node as, for example, [2001:0db8:3c4d::abcd] in brackets in the interface, otherwise I get a 500 error), WebODM cannot communicate with the NodeODM server (no internal IPv6 connectivity).

In the current Nginx configuration, the server does not listen on the IPv6 layer. By default, it binds only to 0.0.0.0:8000 (IPv4), which prevents it from accepting IPv6 connections, even when Docker exposes the port on [::]:8000. This limitation blocks communication with IPv6-only nodes and restricts the ability to fully leverage IPv6 in WebODM deployments.

### DIFF
--- a/docker-compose.ipv6.yml
+++ b/docker-compose.ipv6.yml
@@ -1,0 +1,9 @@
+version: '2.2'
+networks:
+  default:
+    enable_ipv6: true
+    ipam:
+      driver: default
+      config:
+        - subnet: "172.21.0.0/16"
+        - subnet: "fdec:cc68:5178::/64"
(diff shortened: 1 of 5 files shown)


pierotofy:
Looks good to me, thank you @as598466 !

PR #1641, opened 2025-03-31
Docker compose deprecated "version" attribute + latest Tag + update config summary

as598466:
Hello!

I was constantly getting warning messages about the "version" attribute being obsolete in the latest version of Docker Compose. I’ve removed it from all the Docker Compose files. While it’s not a major blocking issue, this makes the script more aligned with the latest versions. 
Before :
Capture d'écran 2025-03-31 115543
After :
Capture d'écran 2025-03-31 130557

I’ve also specified image: opendronemap/imagename:latest with the "latest" tag, following official recommendations.

Added to the configuration summary the value IPV6: YES if it’s being used.

Capture d'écran 2025-03-31 130731

+ Update Redis to 7.0.15 (previous 7.0.10)

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -1,6 +1,6 @@
 # This configuration builds from Dockerfiles 
 # instead of pulling from opendronemap hub
-version: '2.1'
+
 services:
   db:
     build: ./db
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,4 +1,3 @@
-version: '2.1'
 services:
   webapp:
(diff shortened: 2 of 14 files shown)


pierotofy:
Thanks @as598466, as far as the deprecation notices, see our discussion here: https://github.com/OpenDroneMap/WebODM/pull/1570

I just don't know if removing it would break older versions.

as598466:
Thank you for your feedback!
This only concerns older versions of docker-compose. Aren’t people supposed to update their configuration ?
I’ve just tested it on Debian 12 with the latest version of Docker and Docker Compose, and it worked without any issues. Maybe it’d be a good idea to include a list of supported systems in the official documentation (e.g., Debian 11, 12, Ubuntu 20.04, 22.04, etc.) to encourage users to keep their configurations up to date. Just a simple suggestion!

pierotofy:
> Aren’t people supposed to update their configuration ?

It would be really nice if they did :)

PR #1645, opened 2025-04-01
Update webodm.sh

as598466:
* Add IPv6 summary configuration

Capture d'écran 2025-03-31 130731

* Remove useless (stop, down)  if [[ $ipv6 = true ]]; then command+=" -f docker-compose.ipv6.yml" fi


pierotofy:
Excellent, thank you!!

PR #1653, opened 2025-04-13
Plugin enable/disable "fix"

tenchirocom:
This affects one file, app/plugins/functions.py. Two methods in this file are touched: enable_plugin() and disable_plugin().

The original methods attempt to call enable()/disable() methods on plugins. It appears these methods can sometimes be called even if a plugin is enabled or disabled. Further, these methods are not currently included in the plugin_base.py file. Presently, this just throws an exception which is caught later. This pull request cleans up these functions a bit. It ensures, by checking the database enabled flag, that the methods need to be called. It also checks to ensure the plugin has an enable/disable method. It distinguishes between no method present, and an actual failure of execution of an existing method.

If there is no method, then it returns cleanly and sets or clears the database flag appropriately. However, if the method is present and fails, then enable_plugin() prints a warning, and raises an exception WITHOUT setting the db enable flag. ASSUMPTION: enable failed, and thus the plugin is not functioning correctly, thus the UI should show it is not running. disable_plugin() responds differently, if it fails, it logs the warning, but does not raise an exception. ASSUMPTION: the UI can remove the plugin from the menu whether or not it was able to disable cleanly. Thus the db enabled flag is returned to false.

ASSUMPTION: existing plugins might not expect the original behavior. It's possible this might break plugins that depend on enable/disable running automatically on startup, as a kind of __init__ alternative. Scanning the current coreplugins, I don't see any actually using this method, or the plugin enabled flag, but it is possible a 3rd party might have.

Why it's needed: A recent plugin I was developing to install and uninstall a custom brand, info, and them required these to properly install and uninstall static resources like icons and backgrounds. They seemed to be called when the plugin had already been enabled or disabled, like on startup. But this behavior overlaps with __init__, at least for startup enablement. The plugin itself then needs to check the database enabled flag, which is not necessary if the semantics are more strict, or alternatively create a new database flag to represent specific plugin enablement, but this seems redundant. Thus this proposed change.

An additional modification might be to add abstract classes for enable/disable in PluginBase as well as the class level enabled flag to make it clearer what to do with these.


pierotofy:
I'm not sure I understand what is being fixed here. Are you trying to add enable/disable functions to PluginBase because your plugin needs to know when it's being enabled? https://github.com/OpenDroneMap/WebODM/blob/master/app/plugins/plugin_base.py

tenchirocom:
Yes. And it wasn't working right because functions.py calls an enable/disable, but it is not called ONLY when the button is pressed. There is sometime configuration work that *only* needs do be done once, on enablement. That is, when the admin pushes the enable button.

The original functions.py seems to try to provide this. These functions, enable_plugin() and disable_plugin(), already exist. But it is sometimes called on initialization. I.e. not just when the button is pressed. Also, these methods are generally non-existent. This plugin_base.py does not provide them as far as I can see, unless I'm missing something. So the current code is just throwing exceptions, which are caught one call above. Since I don't find enable/disable in plugin_base (which you reference), or being used anywhere, it seems like maybe it was once used, then forgotten and atrophied?

The change just ensures they are called only when the buttons are pressed, and then do a bit more error handling.

pierotofy:
> So the current code is just throwing exceptions

That's the part I don't understand; can you provide a test case that allows us to reproduce what you are seeing?

tenchirocom:
Thanks for the responses. Sure. I'll see what I can do.

tenchirocom:
After reviewing the specifics, my explanation was indeed unclear. Thank you for drawing attention to that. The exceptions do not come from the existing code. You are correct, it is functioning properly.

The exceptions come from the addition of the p.enable() line (and p.disable) to invoke the new plugin methods. Thus, this request is really about adding a new plugin feature to invoke methods on enable and disable. Apologies for the confusion.

Option 1: Extend register with unregister. The register() method might be used for enable. But there does not seem to be a corresponding unregister() method to clean up. Plugins could just override register() and an unregister() can be added to plugin_base, with corresponding call in disable_plugin. The register() base class method is doing work, so the plugin developer will have to invoke the base class method explicitly if this is to be preserved.

Option 2: Add new enable/disable methods to plugin_base.py and invoke them in enable/disable_plugin().  I can clean up the implementation so it simply adds the try code to invoke the methods, checking for the enable attribute for compatibility. The enable/disable method names match the UI terminology a little closer. But of course, register is already there.

Option 3: Do nothing. Personally, I think the clean-up is important when complex plugins are modifying the underlying system. For example, on disable, my plugin restores the themes and brands to their original state, which was saved on enable. I'm not sure how to do that w/o a disable or unregister. I'd probably have to add a new field to the db like "installed" then check on init and if it falls out of sync with enabled, do the work and sync. But this seems inelegant.

In the end, this does seem to be an enhancement not a "fix" as I originally suggested. If you are interested to entertain it and have a preference of which option, I will rework the request and resubmit. Any option seems would not affect existing plugins in any way.

pierotofy:
> Option 2: Add new enable/disable methods to plugin_base.py and invoke them in enable/disable_plugin()

This seems like the way to go in my opinion. If you can open a new PR or rework the existing one we can certainly add this. 🙏

PR #1659, opened 2025-04-22
Pan support, mobile support for zoom/pan

pierotofy:
Continuation of #1658, adds mobile support, pan support, better zoom (makes touchpads with "acceleration" scroll way smoother instead of using a fixed zoom factor).


pierotofy:
Thought I'd never have to write touch gesture handling code for pan/zoom again after https://github.com/pierotofy/glassomium/blob/master/src/GestureManager.cpp#L113 but here we are :laughing:

PR #1669, opened 2025-05-11
Configurable worker count. Closes #1668.

### LINKED ISSUE #1668: Use available rather than physically present CPU for worker count
WebODM currently defaults to using `2 * <physically_present_hyperthreading_core> + 1` for the gunicorn worker count, based on Linux's `/proc/cpuinfo`, as defined [here in start.sh](https://github.com/OpenDroneMap/WebODM/blob/27ec8a75e51d52f618c73a31da9301c64821ecd8/start.sh#L150).

However, as WebODM more often runs in a containerised environment, it's likely that the amount of actually usable CPU's is lower than the physically present ones.

As gunicorn is based on a pre-fork worker model, each worker starts an independent process, significantly increasing memory pressure for resource constrained environments (like Raspberry Pi's, or FWIW, my ancient home server), while most likely not adding performance. For example, on my TrueNAS Scale server, WebODM is using ~650 MB of RAM while _not doing anything_.

Lastly, the current method is OS-dependent and, hence, would prevent WebODM from running on non-Linux platforms. A commonly available method which is not OS-dependent would be GNU coreutil's `nproc` (which incidentally is available as part of Ubuntu 21.04 which is used to build WebODM containers).

dokterbob:
* Use common `WEB_CONCURRENCY` env variable to configure gunicorn workers.
* Default to using available (rather than physically present) CPU cores, through commonly available and OS-agnostic`nproc`.
* Allow overriding gunicorn workers.

Closes #1668, and then some. 

Rationale; imagine running WebODM as a Docker container with access to 2 cores on a 64 core machine in a hosting farm/cluster. Without this patch, it will spin up `64*2+1` processes causing significant resource usage causing significant resource use at no performance gain. With this patch, it would spawn `2*2+1 = 5` processes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,5 +55,6 @@ services:
       - WO_BROKER
       - WO_DEBUG
       - WO_SECRET_KEY
+      - WEB_CONCURRENCY
     restart: unless-stopped
     oom_score_adj: 250
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,6 +34,7 @@ services:
       - WO_DEV
       - WO_DEV_WATCH_PLUGINS
       - WO_SECRET_KEY
+      - WEB_CONCURRENCY
     restart: unless-stopped
     oom_score_adj: 0
   broker:
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ djangorestframework-guardian==0.3.0
 drf-nested-routers==0.11.1
 funcsigs==1.0.2
 futures==3.1.1
-gunicorn==19.8.0
+gunicorn==23.0.0
 geodeep==0.9.8
 itypes==1.1.0
 kombu==4.6.7
--- a/start.sh
+++ b/start.sh
@@ -159,7 +159,7 @@ else
     congrats
 
     nginx -c $(pwd)/nginx/$conf
-    gunicorn webodm.wsgi --bind unix:/tmp/gunicorn.sock --timeout 300000 --max-requests 500 --workers $((2*$(grep -c '^processor' /proc/cpuinfo)+1)) --preload
+    gunicorn webodm.wsgi --bind unix:/tmp/gunicorn.sock --timeout 300000 --max-requests 500 --workers $WEB_CONCURRENCY --preload
 fi
 
 # If this is executed, it means the previous command failed, don't display the congratulations message
--- a/start.sh
+++ b/start.sh
@@ -144,6 +144,18 @@ else
         conf="nginx-ssl.conf"
     fi
 
+    # Only set if `WEB_CONCURRENCY` is not defined, allows overriding.
+    # See: https://docs.gunicorn.org/en/latest/settings.html#workers
+    if [ -z "$WEB_CONCURRENCY" ]; then
+        # Default gunicorn worker count to 2 x availabe cores + 1
+        # nproc gives _available_ (rather than physicall present) CPU cores.
+        CPU_COUNT=$(nproc)
+        export WEB_CONCURRENCY=$((2*$CPU_COUNT+1))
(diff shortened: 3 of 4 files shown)


pierotofy:
Hey thanks for the PR @dokterbob 🙏 

We should definitely use `nproc`, I wasn't aware that `/proc/cpuinfo` didn't respect the virtual environment settings.

Did you test this?

dokterbob:
Just started doing tests and initial results are disappointing:

1. `nproc` is more generic etc. but it does not correctly detect available CPUs in cgroups -- getting to that is rather hairy and tricky. I would suggest we tell users not to use that.
2. `WEB_CONCURRENCY` was 'only' [introduced to gunicorn in 2021](https://docs.gunicorn.org/en/20.1.0/news.html#id1). WebODM is frozen at `19.8.0`.

Are there specific reasons not to upgrade to the latest version? From the codebase I don't see obvious deep bindings.

pierotofy:
How about adding a `--web-workers` flag to `webodm.sh` and allow the user to set the desired number of workers?

dokterbob:
### `--web-workers` flag
> How about adding a `--web-workers` flag to `webodm.sh` and allow the user to set the desired number of workers?

Makes sense. Would you like that in the current PR or rather as a subsequence separate one? (I would suggest the latter.)

### Why not `--workers`
Yes, `--workers` can be used but I'd assume gunicorn picked their default to `WEB_CONCURRENCY` for a reason: it is commonly used by a lot of container-hosters, so that WebODM would adopt sane defaults.

One options could be not (yet) upgrading gunicorn but still anticipating this default by implementing it in the shell script.

### gunicorn upgrade
> Without needing to upgrade gunicorn, you can just use the `--workers` flag like we already do. https://github.com/OpenDroneMap/WebODM/blob/master/start.sh#L150

We can but _at some point_ it makes sense to upgrade deps. At this point this supposedly public-facing code has an amount of unmaintained dependencies which is worrying and arguably makes it not something viable to run online at, from a compliance and, also, actual security viewpoint.

Hence my question: why not now?

Saijin-Naib:
Are you able to process a known-good dataset from ODMData as well? 

https://github.com/OpenDroneMap/ODMData

dokterbob:
> Are you able to process a known-good dataset from ODMData as well?
> 
> https://github.com/OpenDroneMap/ODMData

The only dataset my local compute can handle is the bananas. ;)

I'm happy to test it through lightning if credits are provided.

Saijin-Naib:
Brighton Beach Road at Defaults without Resizing completes locally on my N3450 with 4GB RAM and 4GB swap on 64GB eMMC. Takes about 47min due to all the swapping 🤣

dokterbob:
> Brighton Beach Road at Defaults without Resizing completes locally on my N3450 with 4GB RAM and 4GB swap on 64GB eMMC. Takes about 47min due to all the swapping 🤣

Thanks for the hint. Can look at that tomorrow.

Saijin-Naib:
If you need more headroom, ORB will complete this dataset nicely with under 2GB peak usage.

pierotofy:
> I chose to lower it to 0 and that works just fine.

I'm sure it works just fine, but I'm not familiar with Celery internals and I'm worried this could affect startup/teardown times (if there are no jobs to process, does Celery destroy the processes and needs to start them back-up again?). Without answering this question I wouldn't change it.

dokterbob:
> > I chose to lower it to 0 and that works just fine.
> 
> I'm sure it works just fine, but I'm not familiar with Celery internals and I'm worried this could affect startup/teardown times (if there are no jobs to process, does Celery destroy the processes and needs to start them back-up again?). Without answering this question I wouldn't change it.

That would be my expectation. The same lag would be expected as when scaling out workers (seconds, I suppose) -- on a always-on production machine it might make sense to disable autoscaling for this reason.

> > Would you like that in the current PR or rather as a subsequence separate one?

> Can be separate.

Created issues for tracking. I suggest `--background-workers` for Celery as the name celery is more of an implementation detail.

PR #1670, opened 2025-05-11
Overhaul/cleanup of main Dockerfile

dokterbob:
The Dockerfile was really hard to maintain with lot's of repetition and unreadable lines, plus it did not allow _any_ form of caching. The PR in it's current state is not cleaned up as to a allow feedback and/or opting out on particular components of it. Happy to rebase if and when consensus is reached.

This PR is a first step towards achieving a more maintainable Docker file, it:
1. Does global configuration first.
2. Caches static/system dependencies which don't depend on files in the repo.
3. Caches Python and Node dependencies based on only relevant files (`requirements.txt`, `package.json`).

There is still much space for improvement, like:
1. Separate slim containers for nginx, celery worker, nodeodm and cron. Basically, one process per container.
2. Multi-stage build where build-time deps are installed only for building... deps.
3. Building off or using standard postgis or postgres container rather than building our own. Though multi-arch support might be an issue here.
4. Enabling efficient caching for NodeODM, webpack, Django static files, plugins and translations by clarifying/changing project structure.


pierotofy:
Thanks for the PR; the reason why we don't have a "neat" Dockerfile is because docker creates a new layer for *every* RUN command. This speeds up build time but increases the size of the image drastically.

I would think that using multi-stage builds is the recommended way to go currently https://docs.docker.com/build/building/multi-stage/ but I haven't looked into it. 

We cannot use the official postgres/postgis images, see https://github.com/OpenDroneMap/WebODM/issues/826

dokterbob:
Thanks for the quick feedback!

TLDR; happy to work on multi-stage build, need to know which deps are run and which are build-time and some help with testing end result.

I completely understand that you don't want to merge this as-is. I'd love to see what's necessary to get some improvements underway though.

> Thanks for the PR; the reason why we don't have a "neat" Dockerfile is because docker creates a new layer for _every_ RUN command. This speeds up build time but increases the size of the image drastically. I cannot merge this PR as-is.

It does create a layer for every RUN, but AFAIK that doesn't necessarily increase image size (because it's all stuff we'll need later on), other than that build requirements are now part of these run-time layers (e.g. what's removed [here](https://github.com/OpenDroneMap/WebODM/pull/1670/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R59)). 

As suggested, a multi-stage build would take care of that --- but it's a bit tricky to see clearly which deps (e.g. fs packages) are necessary and which not. I'd love guidance and help testing on that and I can roll forward with a multi-stage build, which would yield slim run-time images and fat build-time images, both with efficient caching.

Regarding postgres, I see your point regarding upgrades for non-technical users. It would require a one-off dump before and a load after the upgrade which, even for somewhat technical users isn't exactly a walk in the park. Let me ponder this for a bit. ;)

Saijin-Naib:
Do you think we could replicate what is used in PGSQL in SQLite3/geopackage and avoid the PGSQL dependency and import/export entirely?

dokterbob:
> Do you think we could replicate what is used in PGSQL in SQLite3/geopackage and avoid the PGSQL dependency and import/export entirely?

Most probably, in theory, we could. But there are so many stale dependencies that I think we have other problems. Postgis is way better anyways and not at all that heavy (considering the weight of ODM itself).

On a different note, indeed, layers are heavier than expected. I've already chunked together many steps but still seeing much more weight than expected. Looking into this now.
image

Saijin-Naib:
One consideration I have aside from build/package complexity with PGSQL, is total size of the docker images to download from a fresh start. 

We have a lot of users on highly bandwidth-constrained connections, so thinking about whether or not a few hundred MBs can be dropped by moving to a library we already use elsewhere (even if technically less better, so long as it does the job well) is something I have been focusing a lot on.

If you have further thoughts in that vein as you poke around more, I would love to hear them. 

It sounds like your multistage reorg may address that a bit?

dokterbob:
> One consideration I have aside from build/package complexity with PGSQL, is total size of the docker images to download from a fresh start.
> 
> We have a lot of users on highly bandwidth-constrained connections, so thinking about whether or not a few hundred MBs can be dropped by moving to a library we already use elsewhere (even if technically less better, so long as it does the job well) is something I have been focusing a lot on.
> 
> If you have further thoughts in that vein as you poke around more, I would love to hear them.
> 
> It sounds like your multistage reorg may address that a bit?

Some of it, yes. Right now I am breaking my brains over the fact that my docker image ends up being 1GB+ while the original one was 650 MB. Normally, images like these are <300MB, there's probably room for improvement.

As just one example: we're installing and packages both Python 2 and Python 3. Where are we using Python 2? I don't see it being called anywhere in WebODM, but there has to have been a reason it was included, once and perhaps still now.

Regardless, this will be a tedious and piecemeal process!

dokterbob:
image

Even if I lump all steps into one, I still get a huge image!

dokterbob:
> image
> 
> Even if I lump all steps into one, I still get a huge image!
> 
> I'm going to try from master to see whether I manage to get to the same result as what's on Dockerhub.
> 
> The reason I started this is exactly the fact that it takes minutes to rebuild docker images, making optimising and testing them considerably harder.

It does indeed seem that my Docker build from `master` _has executed the exact same steps_ yet came out at 1.77 GB versus 680 MB on Dockerhub. 
image

Perhaps this is just that one is compressed or perhaps something changed since 2.8.1's release. Doing a build of the latter now to confirm. If that is not the case then I guess it's compression and the bloat in my own image wasn't as extensive as I thought (and, consequently, some small cleanup is sufficient).

dokterbob:
@pierotofy Just did a lot of cleanup, comparison and local testing:

1. Consolidated layers shrinking images size in this PR down from ~2.1 GB to 1.89 GB versus 1.77 GB for current production images. Compressed that's <xxx> versus <yyy> before. Significant gains can be made here, but not sure about the timeline. Given the cleanup and caching I would already consider this a _significant_ improvement -- custom builds now run in about 1m versus 8m before.
2. Several fixups after local testing -- and as such confirmed to work (including #1669).

dokterbob:
FYI I'm locally experimenting with an Ubuntu 22.04-based multi-stage build (staying with Python 3.9 and Node 20).

I got bind layer caching working (so we don't have to download deps again all the time) and the image size is currently at 1.48 GB. Not yet the whopping savings I'd have hoped for. But effectively, as we're using the much more common base image should yield significant savings.

Opened a WIP PR here: https://github.com/OpenDroneMap/WebODM/pull/1672
There be dragons. 🐉

Saijin-Naib:
Feel like going wild and trying Alpine base 🤣

That has been a long-term goal for me to save a ton of space

docker-compose.ssl-manual.yml
```diff
@@ -1,5 +1,4 @@
 # This configuration adds the volumes necessary for SSL manual setup
-version: '2.1'
```
pierotofy, line 2: See the discussion here for this: https://github.com/OpenDroneMap/WebODM/pull/1570

sources.list
```diff
@@ -0,0 +1,8 @@
+# Old-releases for 21.04
```
pierotofy, line 1: Could we keep the sources list inside the Dockerfile?

dokterbob:
> Feel like going wild and trying Alpine base 🤣

🥳 

However, Alpine uses Muscl which is a different beast from libc. Ideally, you want to use the same base image across all ODM images and for compute-heavy stuff muscl might be risky. First step AFAIK is figuring out exactly which are run and which are build-time deps.

Then, for multistage, testing testing testing testing (manual!).

> That has been a long-term goal for me to save a ton of space

First, let's see what deps are what!

> This looks really great!

Appreciate the quick and positive feedback!

docker-compose.ssl-manual.yml
dokterbob: Seen. Docker-compose's support ended in 2023. https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/

Also, we def. need v2 format to support things like multi-stage.

dokterbob: Note that sticking with v1 I'm not sure if it makes sense for me to actively contribute here, I might fork to be able to move ahead.

pierotofy: Aren't we on v2.1 currently?

My main concern (from issue #1570) is whether removing the version number definition will break the software for people still on some old version of docker-compose.

dokterbob: Honestly, I'd rather not go for latest - I'd aim for supported LTS releases of everything.

Running unpatched software on the open internet is not something I'd like to do even on my home server and will not pass any form of compliance requirements.

What I am saying is, in it's current state, WebODM _is not safe for public use_.

So that's what I'd be changing, step by step.

I'm happy to think about supporting users on upgrade paths though. As there's already import/export functionality, perhaps it can be extended to do mass import/export?

Is there perhaps a way for us to have an in-sync video call or even chat to see if we can align on these issues? I'd be happy to put a considerable amount of time into it over the coming weeks.

pierotofy:
Looks good to me, thanks @dokterbob !

For discussions about goals and future developments, I'd prefer to keep conversations public. Instead of a video call, feel free to open a discussion on https://community.opendronemap.org/c/developers-chat/21

PR #1682, opened 2025-05-19
Allow selectively testing (parts) front/backend with arguments

dokterbob:
Allow selective running of frontend and/or backend tests as well as passing arbitrary arguments to test commands, e.g. to run selective (failing) tests faster.

Tested to pass arguments through to the same command in Docker.

Example: `./webodm.sh test backend app.tests.TestApi`

(Not sure this is the exact path of the test, but the idea should be clear.)

### DIFF
--- a/webodm.sh
+++ b/webodm.sh
@@ -528,6 +528,7 @@ run_tests(){
             echo -e "\033[1mDone!\033[0m Everything looks in order."
         fi
     else
+		environment_check
         echo "Running tests in webapp container"
         run "$docker_compose exec webapp /bin/bash -c \"/webodm/webodm.sh test $@\""
     fi
--- a/webodm.sh
+++ b/webodm.sh
@@ -174,7 +174,7 @@ usage(){
   echo "	liveupdate		Update WebODM to the latest release without stopping it"
   echo "	rebuild			Rebuild all docker containers and perform cleanups"
   echo "	checkenv		Do an environment check and install missing components"
-  echo "	test			Run the unit test suite (developers only)"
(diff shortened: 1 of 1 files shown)


webodm.sh
```diff
@@ -510,17 +510,26 @@ run_tests(){
...
+        fi
+
+        if [[ $test_type = "all" ]]; then
+            echo ""
+            echo -e "\033[1mDone!\033[0m Everything looks in order."
+        fi
     else
         echo "Running tests in webapp container"
-        run "$docker_compose exec webapp /bin/bash -c \"/webodm/webodm.sh test\""
+        run "$docker_compose exec webapp /bin/bash -c \"/webodm/webodm.sh test $@\""
```
pierotofy, line 532: There seems to be some problem with the expansion of the $@ variable here:

```bash
[pt:~/Documents/WebODM] selective_tests(+2/-1) 1 ± ./webodm.sh test all app.tests.TestApi
Running tests in webapp container
docker-compose exec webapp /bin/bash -c "/webodm/webodm.sh test all
./webodm.sh: eval: line 348: unexpected EOF while looking for matching `"'
```

pierotofy: [pt:~/Documents/WebODM_clean] selective_tests(+1/-1) ± bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)
Copyright (C) 2007 Free Software Foundation, Inc.

dokterbob: Replicated it and (hopefully) solved.

pierotofy:
Looks good, thanks!

PR #1687, opened 2025-05-21
Extra work from linter ;)

dokterbob:
Backport of assorted (manual) cleanup and fixes based on feedback from linter (ruff), without actually configuring said linter in ODM repo. Original work: https://github.com/treescape-io/WebODM/pull/1

Consider this a mix-and-match PR, let me know if there's anything you would not like and I'll happily remove it.

### DIFF
--- a/coreplugins/cloudimport/api_views.py
+++ b/coreplugins/cloudimport/api_views.py
@@ -6,7 +6,8 @@ from pyodm.exceptions import NodeServerError
 from app import models, pending_actions
 from app.plugins.views import TaskView
 from app.plugins.worker import run_function_async
-from app.plugins import get_current_plugin
+from app.plugins import get_current_plugin, logger
+from app.security import path_traversal_check
 
 from worker.celery import app
 from rest_framework.response import Response
@@ -100,11 +101,6 @@ class PlatformsTaskView(TaskView):
 
 
 def import_files(task_id, files):
-    import requests
-    from app import models
-    from app.plugins import logger
-    from app.security import path_traversal_check
-
     def download_file(task, file):
         path = path_traversal_check(task.task_path(file['name']), task.task_path())
         download_stream = requests.get(file['url'], stream=True, timeout=60)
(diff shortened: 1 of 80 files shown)


coreplugins/cloudimport/api_views.py
```diff
@@ -100,11 +101,6 @@ def get(self, request):
 
 
 def import_files(task_id, files):
-    import requests
```
pierotofy, line 103: This is exactly why I dislike linters. You're breaking this function because it's executed in an async context, where imports are **not** available from the top module level. But the linter can't possibly know this and suggests to move the imports at the top.

pierotofy:
I'm OK fixing errors (e.g. fixing a reference to `valueError` instead of the correct `ValueError`), but I'm not OK merging the work of a linter because a) I can't trust it b) it's difficult to review by hand and c) Some of the changes are just a matter of syntax and just make it difficult to trace changes across git. (e.g. ` not x in arr` vs `x not in arr`).

dokterbob:
> I'm OK fixing errors (e.g. fixing a reference to `valueError` instead of the correct `ValueError`), but I'm not OK merging the work of a linter because a) I can't trust it b) it's difficult to review by hand and c) Some of the changes are just a matter of syntax and just make it difficult to trace changes across git. (e.g. ` not x in arr` vs `x not in arr`).

Appreciate the feedback and the push-back is not unexpected.

It's why I took care to group various types of changes in commits. However, not knowing the codebase well, it's a bit tricky for me to know which are errors and which are not. 

Perhaps could suggest which ones are obvious fixes and I'll shoot a new PR?
Alternatively, I could suggest a subset of commits myself.

pierotofy:
I think we should definitely try to fix typos that are errors (e.g. `valueError` instead of `ValueError`). 

I don't think we should make changes that are purely stylistic, unless there's a good reason besides "make my linter happy" or "it doesn't respect PEP X".

PR #1694, opened 2025-06-20
Add demo / auto login to WebODM

smathermather:


pierotofy:
This branch should not be merged as-is. Moving to https://github.com/OpenDroneMap/WebODM/issues/1709

Issue #1698, opened 2025-06-26
Docker Compose for Service


Blu3wolf:
> If you wish to run the docker version with auto start/monitoring/stop, etc, as a systemd style Linux Service, a systemd unit file is included in the service folder of the repo.
>
> The following pre-requisites are required:
>
>    Requires odm user
>    Requires docker installed via system (ubuntu: sudo apt-get install docker.io)
>    Requires 'screen' package to be installed
>    Requires odm user member of docker group
>    Required WebODM directory checked out/cloned to /opt/WebODM
>    Requires that /opt/WebODM is recursively owned by odm:odm
>    Requires that a Python 3 environment is used at /opt/WebODM/python3-venv

Couldn't all this be moved into the docker container? This is pretty much what docker is built for, is to handle the dependencies. Screen seems like it could be safely replaced with a "docker compose up -d" for example. Same goes for the Python environment, wouldn't all that be better off inside a container?

Or am I missing something Docker can't or shouldn't do?

pierotofy:
Could we move this conversation over to the forum at https://community.opendronemap.org? :pray: The forum is the right place to ask questions (we try to keep the GitHub issue tracker for feature requests and bugs only). Thank you! :+1:

smathermather:
@Blu3wolf is correct. This isn't the way to run WebODM as a service. I'll remove from the readme.

PR #1715, opened 2025-07-25
Remove systemd write up

smathermather:
Systemd write up only tested on older linux, probably not the best way to keep running now. Removed from README closing #1698


Saijin-Naib:
At least on Alpine with an OpenRC system service, any running Docker containers are automatically restarted when the system restarts, provided the Docker OpenRC service is installed in a runlevel. 

So, I would expect recent Docker under systemd would behave similarly.

PR #1745, opened 2025-09-17
Support for safe textured models

pierotofy:
This PR adds support for ~~progressive~~ safe meshes, which allows to:
 - Prevent crashes when loading large textured models in the browser by loading reduced/simplified versions of the 3D textured model
 - ~~Greatly speed up time to first render by loading progressive LOD models~~


pierotofy:
Unfortunately we won't have full-on progressive meshes, as after testing and even adding KTX2 compression (https://github.com/pierotofy/WebODM/tree/ktx2), swapping LOD in real time causes the browser to freeze rendering, which is an unacceptable tradeoff, so I will simply add the ability to load a safe version of the GLB model which does not crash the browser in cases where models are too large to be displayed. :unamused: 

Future approaches might require implementing something like better 3D tiles support.

Data-Delivered:
Thanks for the effort.  I would absolutely love LOD to work on meshes in WebODM.  Is there a world in which CesiumJS can be incorporated to the ecosystem to achieve this.  I’m no programmer so I have no clue, but potree for Point Clouds and CesiumJS for mesh models would be a sick combo!

pierotofy:
CesiumJS is a great library, but tries to do too much for WebODM's needs. 3D tiles could be loaded directly in THREE.js

pierotofy:
chrome-capture-2025-09-19

Data-Delivered:
Awesome! Well done!Kevin ***@***.*** cell/textOn Sep 19, 2025, at 1:45 PM, Piero Toffanin ***@***.***> wrote:﻿pierotofy left a comment (OpenDroneMap/WebODM#1745)
chrome-capture-2025-09-19.gif (view on web)

Data-Delivered:
Good to know. I may be experimenting this winter with both of these for our in house needs. Thanks for all you do!  Kevin ***@***.*** cell/textOn Sep 19, 2025, at 1:41 PM, Piero Toffanin ***@***.***> wrote:﻿pierotofy left a comment (OpenDroneMap/WebODM#1745)
CesiumJS is a great library, but tries to do too much for WebODM's needs. 3D tiles could be loaded directly in THREE.js

PR #1752, opened 2025-09-24
Add comments to Units.js

dokterbob:
For reasons(tm), I needed to get a proper grip on the functionality in `Units.js`. I asked a friendly neighbor AI to add comments and reviewed them manually. It also fixed up some minor formatting and indentation issues.

To the best of my knowledge, the comments accurately describe functionality and, in the process, even brought up two classes of bugs:
1. Missing `new` keyword when returning `NanUnit()` (5 places)
2. Wrong variable names in `toMetric`: should be `units.sqmeters` and `units.cbmeters` instead of `unit.sqmeters` and `unit.cbmeters`

In JavaScript, when you define something with the class keyword, you must use new to instantiate it. If you try to call a class constructor without new, JavaScript will throw a TypeError. [Ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes#constructing_a_class)

_After_ this PR I'm happy to shoot a 2nd PR for aforementioned bugs, or I could add them here, as you please. :)

PR #1753, opened 2025-09-25
Minor fixes in Units.js

dokterbob:
AI discovered, human fixed. Ref: #1752 

Commit messages should be sufficiently clear.


pierotofy:
Thanks @dokterbob !

Issue #1300, opened 2023-03-14
Fix docs build


pierotofy:
The docs build is currently broken.
```
Git error: command `git clone --bare --no-hardlinks --quiet --no-tags --depth 1
--single-branch --branch 4.x -- https://github.com/middleman/middleman.git
/opt/hostedtoolcache/Ruby/2.7.7/x64/lib/ruby/gems/2.7.0/cache/bundler/git/middleman-0aa2a412fad4ddc12db204bb84a86fee137bc28e`
has failed.
Revision 4.x does not exist in the repository
https://github.com/middleman/middleman.git. Maybe you misspelled it?
Error: Process completed with exit code [11](https://github.com/OpenDroneMap/WebODM/actions/runs/4418839955/jobs/7746487447#step:4:12).
```

chris-bateman:
@pierotofy Would this potentially explain why some doco is missing?
https://github.com/OpenDroneMap/WebODM/blob/6bc90025aada533b38af7181cebc29685e0c9af4/app/api/tasks.py#L172

Is there an expectation that I should be able to see this in the doco?
http://docs.webodm.org/#task

pierotofy:
Still broken. 

```
Run bundle exec middleman build
bundler: failed to load command: middleman (/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/middleman)
/opt/hostedtoolcache/Ruby/2.7.8/x64/lib/ruby/gems/2.7.0/gems/middleman-syntax-3.2.0/lib/middleman-syntax/haml_monkey_patch.rb:4:in `<module:Haml>': Filters is not a module (TypeError)
/opt/hostedtoolcache/Ruby/2.7.8/x64/lib/ruby/gems/2.7.0/gems/haml-6.3.0/lib/haml/filters/base.rb:5: previous definition of Filters was here
	from /opt/hostedtoolcache/Ruby/2.7.8/x64/lib/ruby/gems/2.7.0/gems/middleman-syntax-3.2.0/lib/middleman-syntax/haml_monkey_patch.rb:3:in `<top (required)>'
	from /opt/hostedtoolcache/Ruby/2.7.8/x64/lib/ruby/gems/2.7.0/gems/middleman-syntax-3.2.0/lib/middleman-syntax/extension.rb:4:in `require'
[...]

PR #1765, opened 2025-10-12
Update docs

pierotofy:
Replaces slate with astro, which adds:
 * More modern interface
 * Client side full text search
 * Light/dark mode

Updated a few guidelines. Lots of information is still out of date though.


pierotofy:
Closes #1300

PR #1767, opened 2025-10-13
Plugins Developer Guide

### LINKED ISSUE #494: Document Plugin API
As the plugins API starts to stabilize, we need to start documenting it so that people can utilize it. At a minimum links to the example plugins should be provided.

pierotofy:
Closes #494 

Work in progress..


diegoaces:
Thanks, Piero. This contribution will be an excellent help for those who are just starting to develop plugins for WebODM, as well as for those who began by just reading the example code (myself included).

PR #1779, opened 2025-11-10
Commit retries

pierotofy:
Adds support for retrying the last call to file uploads (commit), including adding a 30 second timeout from the origin server. If the commit call hangs, the entire upload fails, so this makes it more robust.


pierotofy:
Also changes the chunk upload sizes from 8MB to 4MB to improve uploads on really bad networks.

Saijin-Naib:
2MB might allow partial uploads of 12MP images (usually 4MB or less), but 4MB chunks should hugely improve uploads on spotty networks for sure.

PR #1784, opened 2025-11-12
Fix getArea()

### LINKED ISSUE #1705: Relato de Erro - Discrepância entre medidas no modelo 3D e na ortofoto (2D)
Hello,

I would like to report a significant discrepancy between the measurements taken from the 3D model and those from the orthophoto (2D) generated by the software.

I understand that terrain elevation can influence measurements, and that measurements based on the orthophoto are naturally limited to two dimensions (2D). However, the difference observed between the 3D and 2D measurements is extremely large—well beyond what would be expected from topographic variation alone.

This discrepancy is compromising the reliability of the measurements for both technical and practical purposes.

I would like to know if this is a known behavior of the software, if there are any recommended best practices to avoid this type of discrepancy, or if it could be the result of a processing error.

Below are screenshots clearly showing the stark difference between the measurements.

Image

Measurements from the orthophoto (2D)

Image

Measurements from the 3D model

If needed, I can provide additional information, files, or any other data that may help investigate the issue. I’m available to assist.

pierotofy:
Potree's `getArea` function is incorrect. This PR fixes that:

Old

image

New

image

Should close #1705

### DIFF
--- a/app/static/app/js/vendor/potree/build/potree/potree.js
+++ b/app/static/app/js/vendor/potree/build/potree/potree.js
@@ -54157,10 +54157,8 @@
 			}
 
 			// The final area is half the magnitude of the resulting area vector.
-			// Magnitude = sqrt(Ax^2 + Ay^2 + Az^2)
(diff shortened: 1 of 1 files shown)


pierotofy:
Also adds a better unit selector and imperial (US) units support.

image

PR #1785, opened 2025-11-14
Support for non-UTM coordinate systems, GCP improvements, 3D model view improvements

pierotofy:
Changes to bring support for non-UTM, non-metric coordinate system outputs (WIP).

Also better support for GCP file workflow, now with client side validation.

image


pierotofy:
image

pierotofy:
Adds a camera size slider too: 

image

image

pierotofy:
Now some tests and this is good to go.

PR #1786, opened 2025-11-17
Update node-odm image to stable version

pierotofy:
The current ODM 24.04 update has problems, which are surfacing in threads like https://community.opendronemap.org/t/project-webodm-works-vs-nodeodm-error/25584/2 and https://github.com/OpenDroneMap/ODM/issues/1949

I'm pinning the NodeODM image to a stable (pre 24.04 update) tag until things are more battle tested.

### DIFF
--- a/docker-compose.nodeodm.yml
+++ b/docker-compose.nodeodm.yml
@@ -10,7 +10,7 @@ services:
     environment:
       - WO_DEFAULT_NODES
   node-odm:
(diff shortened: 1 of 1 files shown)


smathermather:
Full agreed. Hopefully we can get these issues sorted out shortly. I do have someone who may be able to be maintainer, so working out the logistics on that quickly.

PR #1787, opened 2025-11-20
Update eventlet to fix Windows compatibility issues

sylveon:


pierotofy:
What compatibility issue is being fixed?

sylveon:
This one: https://stackoverflow.com/questions/75137717/eventlet-dns-python-attribute-error-module-dns-rdtypes-has-no-attribute-any

PR #1795, opened 2025-11-28
Potree dynamic unit measurements support

pierotofy:
Fixes the display of the proper units when changing between imperial/metric systems in the various Potree measurements.


pierotofy:
image

image

PR #1800, opened 2025-12-06
Add settings.PLUGINS_CORE_IGNORE_SCRIPTS to workaround an issue under the MSIX PSF

sylveon:
When WebODM is packaged in MSIX and the install location is redirected with the file redirection PSF fixup, it is not possible to set the current working directory to be a folder created under the redirection [(source)](https://github.com/microsoft/MSIX-PackageSupportFramework/blob/main/fixups/FileRedirectionFixup/readme.md#changing-directories). `npm` does this when running postinstall scripts. This meant that WebODM failed to initialize some of its core plugins.

Thankfully, the only postinstall script in core plugins appear to be core-js's advertisement script. So we can just add a settings to disable those. User plugins still get their postinstall scripts ran normally, because those are not subject to the PSF.

### DIFF
--- a/app/plugins/functions.py
+++ b/app/plugins/functions.py
@@ -93,7 +93,7 @@ def build_plugins():
                 if platform.system() == "Windows":
                     npm = "npm.cmd"
                 command = [npm, 'install']
-                if settings.PLUGINS_CORE_IGNORE_SCRIPTS and plugin.is_persistent():
+                if plugin.is_persistent():
                     command.append('--ignore-scripts')
                 subprocess.call(command, cwd=plugin.get_path("public"))
             except FileNotFoundError:
--- a/webodm/settings.py
+++ b/webodm/settings.py
@@ -79,10 +79,6 @@ PLUGINS_BLACKLIST = [
     #'measure',
 ]
 
-# Do not run npm postinstall scripts when installing coreplugin dependencies,
-# workaround for core-js's postinstall script breaking under the MSIX PSF.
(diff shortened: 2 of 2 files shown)


pierotofy:
Perhaps we should always pass `--ignore-scripts` regardless of environment? Is there a plugin that does not build correctly with that flag? I'd rather not add additional flags in the settings if possible. 

@Saijin-Naib can we also get more details about the scope of these additions? Like, I get that this has to do with the desktop installer, what exactly is being improved here with the introduction of MSIX that didn't work with Innosetup?

sylveon:
The core plugins appear to work, but since user plugins can be arbitrary, they could depend on post-install scripts (I'd hope they wouldn't). However, always passing `--ignore-scripts` for coreplugins sounds reasonable, I can do that if desired.

sylveon:
As for context, I was contracted to help MSIX-package the WebODM desktop installer.

The main objectives are:
- Easier deployment in managed environment
- Automatically managed updates via the Microsoft Store
- Easier way to move the app & data
- Easier license management (built-in to the Store)
- Hopefully reduce AV detection rates
- Support for free trials provided via the Store

pierotofy:
> always passing --ignore-scripts for coreplugins sounds reasonable

Yes, if you've tested the core plugins and they all work, I don't think we need to worry (too much) about third party ones and it would be preferable (and probably a good security enhancement).

Thanks for the additional details and context. It makes sense.

PR #1816, opened 2025-12-24
Add support for CRSes not in the EPSG database

pierotofy:
Some CRSes do not have EPSG codes, which WebODM does not handle correctly.

This PR adds support for handling CRSes that don't have an explicit EPSG code by storing the georef information as WKT when an EPSG code is not available.


pierotofy:
Also adds CRS information in the task details from the project list, one-click copy to clipboard for task IDs.

image

PR #1818, opened 2025-12-29
Add extra environment debugging information in webodm.sh

wloxyz:
At the request of @Saijin-Naib as a part of a project to get the full stack working on Podman, I'd made some modifications to webodm.sh to assist users in debugging. The `--debug` flag now prints further information about the user's environment, including the value of their DOCKER_HOST environment variable as well as the versions of docker and docker-compose they're using. It also detects some edge-cases that may lead to unintended crashes, and will suggest things the user should look at in case they're affected.

This pull request links to documentation not yet merged in https://github.com/OpenDroneMap/docs/pull/256, so ideally that should be focused on upstreaming first.


pierotofy:
Thanks @wloxyz 🙏 will wait on https://github.com/OpenDroneMap/docs/pull/256 then merge.

Saijin-Naib:
@wloxyz , thanks for getting this in!
@pierotofy , I've merged willow's changes in the documentation and they are live, so once review is done, things should be ready.

I tasked willow with helping to ensure that WebODM can be (as easily as possible) made to work in as many OCI runtimes as possible.

Currently, we use rootful docker, but with their changes above and to the docs, we can more easily support rootless docker, as well as both rootful and rootless podman (and podman desktop where that runs).

This will help us address this infrequent but persistent user request that we can run rootless and/or in podman/podman desktop, and hopefully smooth some edges around first-start for users, as well as users switching between rootful/rootless and/or OCI runtimes.

In terms of usage and testing, I've been dogfooding willow's changes in Alpine under rootless podman and it just simply works. Willow tested on fedora as well as I believe arch (btw), as well.

Podman documentation has been expanding to make setting up rootless under the various distros and flavors way, way more smoother than it was even a year ago, so there really isn't much to figure out for the user other than to know what they are on and how they want to run.

PR #1819, opened 2025-12-31
Fix sharp pixel limit crash for large textures

omwanere:
### What changed
- Disabled Sharp input pixel limit to prevent crashes on large textures

### Why
Sharp throws `Input image exceeds pixel limit` when processing large images,
causing GLB generation to fail.

### Result
Allows large textures to be processed without crashing the pipeline.
Screenshot 2025-12-31 084849

### DIFF
--- a/app/scripts/glbopti.js
+++ b/app/scripts/glbopti.js
@@ -97,7 +97,7 @@ async function main() {
                 encoder,
                 resize: [textureSize, textureSize],
                 targetFormat: undefined,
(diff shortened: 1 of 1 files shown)


pierotofy:
Nice, thanks for the PR @omwanere 🙏 

Can you share a test input GLB that shows the crash before the fix is applied?

Edit: closing assuming this was LLM generated and no fix was actually tested.

Issue #1181, opened 2022-05-07
Possible Bug: PoTree Viewer Does Not Show Labels For Dome Volume Function


Saijin-Naib:
### How did you install WebODM (docker, installer, etc.)?
WebODM Docker/Native

### What's your browser and operating system? (Copy/paste the output of https://www.whatismybrowser.com/)
Alpine Linux Edge (v3.15)

### What is the problem?
The Dome Volume Measurement Tool does not display labels like the Rectangular Prism Volume Measurement Tool does.
![image](https://user-images.githubusercontent.com/19295950/167234253-cb8d9e7d-6edb-4840-87e0-4320aa5efdcb.png)

### How can we reproduce this? (What steps trigger the problem? What parameters are you using for processing? Include screenshots. If you are having issues processing a dataset, you must include a copy of your dataset uploaded on Dropbox, Google Drive or https://dronedb.app)
Use the 3D View to place a Dome Volume Measurement area and verify that no metrics are displayed.

Issue #660, opened 2019-05-02
Plugin Idea: allow users to load already processed orthophotos, DEMs and 3D models


pierotofy:
Seems like a recurring use case for some users.

kikislater:
This function is already present in MicMac for information for DEM. It allows to process, clean dem or point cloud, import it to have a clean Ortho !

smathermather:
@kikislater -- have you seen this?: https://github.com/dronemapper-io/NodeMICMAC

kikislater:
Yes, I have to find time to test it nos ;-)

kikislater:
As we discuss about it, will it be plan in the near future to choose between opensfm and micmac ?

pierotofy:
We'll probably add the option to add a NodeMicMac node when running WebODM, yes.

smathermather:
It makes lots of sense.

pierotofy:
This will likely be included as part of core, as the dependency to PotreeConverter/Entwine is difficult to add as a plugin (for point clouds).

byhkalo:
Hello @pierotofy, @smathermather

I see this issue and this problem is actual for me. I have a lot of orthophotos from Pix4D (I processed up to 1000 images). I would like to import them to present to clients.

How can I support that feature? I don't have experience in the web and can't write code by myself. But can I pay or something like that to support your team. How much you need?

byhkalo:
Hello @pierotofy and @smathermather!
Hope you are going well.
What do you think about that feature? Is it hard to implement?

pierotofy:
Hey @byhkalo :hand: it's not difficult, but just time consuming (making the UI, API endpoints, async processing of point clouds to generate EPTs). Get in touch at https://uav4geo.com/contact we can discuss time/cost estimates.

PR #1838, opened 2026-02-20
Import orthophotos, DEMs, point clouds from external sources

pierotofy:
Work in progress.

image

Also moves the `cloudimport` and `openaerialmap` plugins out to https://github.com/OpenDroneMap/WebODM-Plugins-Archive/

Closes OpenDroneMap/WebODM#660 OpenDroneMap/WebODM-Plugins-Archive#1

Fixes OpenDroneMap/WebODM#1181


pierotofy:
image

pierotofy:
image

smathermather:
Looks like OAM plugin is getting pulled out in this pull request. Do we need a maintainer for that plugin or support from the OAM team otherwise? Looks like based on https://github.com/OpenDroneMap/WebODM-Plugins-Archive/ it's a backend issue, not a plugin issue.

cc @dakotabenjamin, spwoodcock

pierotofy:
Yes, I'm no longer willing to maintain the OAM plugin. It's been broken several times and for that reason I don't think it belongs into the core plugins category. I suggest that OAM makes the plugin available as a downloadable .zip plugin from the OAM website (if there's interest in providing such plugin) which users can download and install into WebODM as needed.

spwoodcock:
Thanks for flagging @smathermather!

I agree with your viewpoint @pierotofy. OAM hasn't seen much attention in a while, so I would understand if the plugin has broken over time.

That said, OAM has since had a bit of a revival. The backend was fully migrated to eoAPI and is based on STAC.
Next up is adding a new uploader API and UI, meaning it will be actively maintained and a lot more functional.

So overall I think it's probably a good thing to move the plugin out of core for now, but would be happy if there was a chance to appeal this in future, after the uploader is in place and there are more guarantees on it's stability 🙏 

p.s. there are **many** users of the plugin currently, so it's a bit of a blow to the community. We need to assess if (1) we tell them to switch to the new uploader website, or (2) ask them to wait a bit of time until the plugin functionality is restored (if an option).

dakotabenjamin:
I should add a specific correction. We don't have exact numbers of uploads to OAM from WebODM sources, but we do have about 1000 datasets with the provider label "WebODM" total, 300 of which uploaded in the last 2 years. My guess is that these are uploads where the user did not specify/change the provider field.

smathermather:
> So overall I think it's probably a good thing to move the plugin out of core for now, but would be happy if there was a chance to appeal this in future, after the uploader is in place and there are more guarantees on it's stability 🙏
> 
> p.s. there are **many** users of the plugin currently, so it's a bit of a blow to the community. We need to assess if (1) we tell them to switch to the new uploader website, or (2) ask them to wait a bit of time until the plugin functionality is restored (if an option).

Now that OAM is back on the mend, maybe we can come up with a strategy for coordinating plugin maintenance on this plugin.  On the question of core, default discoverability is actually quite important for a plugin of this nature. But ensuring there is an identified maintainer and process for maintenance as OAM grows will be important to ensuring the necessary stability, especially with so many users.

PR #1841, opened 2026-02-24
formulas: add BGRNReP and RGBNReP as used by the Micasense Rededge-P

### LINKED ISSUE #1840: automatic band selection in Plant Health view not working for RGBNRe multispectral images
Using images from a multispectral camera like the Micasense RedEdge-P, the "Automatic" band selection uses RGN instead of selecting the RGBNRe. This information is available in the tiff images band names.

This results in wrong colorization of plant health in WebODM

maurerle:
With automatic camera filters, the fallback for NDVI is "RGN" (the first one which contains all bands needed for the calculation). This produces wrong results if the band is not set automatically.

So for the Micasense RedEdge-P, the bands are "BGRNReP" which is currently missing.

Adding this helps to automatically select the camera bands correctly.

fixes #1840 

Dataset for reproduction of the issue:
https://fh-aachen.sciebo.de/s/qMxs3jHH4FKb2S5

### DIFF
--- a/app/api/formulas.py
+++ b/app/api/formulas.py
@@ -152,6 +152,8 @@ camera_filters = [
     'BGRNReL',
     'BGRReNL',
 
+    'BGRNReP',
+    'RGBNReP',
     'RGBNRePL',
 
     'L', # FLIR camera has a single LWIR band
--- a/app/tests/test_formulas.py
+++ b/app/tests/test_formulas.py
@@ -49,31 +49,42 @@ class TestFormulas(TestCase):
             for f in i['filters']:
                 bands = list(set(re.findall(pattern, f)))
                 self.assertTrue(len(bands) <= 5)
-    
+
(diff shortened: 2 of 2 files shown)


pierotofy:
> you didn't include a copy of your dataset for us to reproduce this issue

maurerle:
This is how it looks like wrong - auto is equal to "RGN":

image

The settings used for the processing are:
`auto-boundary:true, dem-resolution:0.5, dsm:true, dtm:true, matcher-neighbors:0, min-num-features:12000, optimize-disk-space:true, orthophoto-resolution:0.5, pc-quality:high, primary-band:red, rerun-from:dataset, sfm-algorithm:planar, smrf-threshold:0.05`

pierotofy:
Looks good, thanks for the patch and the dataset @maurerle 

image

PR #1843, opened 2026-02-26
Export .env variables to child processes

### LINKED ISSUE #1799: variables in .env file not exported by webodm.sh
It's my experience that command line arguments export OK to children processes but variables in the .env file do not.

I fixed it in my environment by changing

source "${__dirname}/.env"

to 

set -a; source "${__dirname}/.env"; set +a

pierotofy:
Closes #1799


pierotofy:
Actually upon more careful scrutiny, this might not be needed.

PR #1851, opened 2026-03-08
Fix RCE vulnerability in plugin worker by replacing eval with importlib

aerofiremapping:
Replace the eval/compile-based async execution pattern with importlib dynamic imports. Instead of serializing raw Python source into Celery messages and eval()ing it on the worker, serialize only the module path and qualified function name, then import and call the function directly.

This eliminates arbitrary code execution risk from crafted Celery queue messages while maintaining full compatibility with all existing plugin callers.

### DIFF
--- a/app/plugins/worker.py
+++ b/app/plugins/worker.py
@@ -1,4 +1,4 @@
-import inspect
+import importlib
 from worker.celery import app
 from webodm import settings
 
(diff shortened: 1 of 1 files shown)


aerofiremapping:
## Summary

- Removes `eval()`/`compile()` usage in `app/plugins/worker.py` that allowed
  arbitrary code execution via crafted Celery queue messages
- Replaces source-code serialization with `importlib`-based dynamic loading —
  only the module path and qualified function name are passed over the queue
- All existing plugin callers (calc_contours, upload_to_ion, import_files,
  calc_volume, detect) are module-level functions and remain fully compatible

## Security Impact

Previously, `eval_async` accepted a raw Python source string via the Celery
message broker. Any actor with queue access could inject and execute arbitrary
code with worker privileges. The new `call_async` task only resolves functions
that exist as importable modules in the codebase.

## Test plan
- [ ] Verify contours plugin generates contours correctly
- [ ] Verify cesiumion upload works end-to-end
- [ ] Verify dronedb import works
- [ ] Verify measure/volume calculation works
- [ ] Verify objdetect plugin runs without errors

pierotofy:
Can you please explain how this RCE can be triggered and provide a concrete example that could be used in a production environment?

On your own, no AI please.

aerofiremapping:
Closing - decided not to proceed with this approach.

Issue #1853, opened 2026-03-15
Bash scripts hardcode Bash path, which doesn't work on NixOS systems


jrddunbr:
### How did you install WebODM (docker, installer, etc.)?

I cloned the repo and used Docker from there (see more below)

### What's your browser and operating system?

Browser: Firefox on Arch Linux

Host system: Headless NixOS (Stable)

### What is the problem?

The bash scripts in the project hard-code the bash path which breaks the scripts on NixOS as `/bin/bash` does not exist.

This is related to (and effectively duplicates) #1110 which appears to have been prematurely closed due to a lack of understanding of NixOS - NixOS breaks many Linux conventions by having an immutable current state filesystem backed by mostly symlinks that are pointing at a configured state of the system that is fully declarative (if you wanted). So, an entire operating system configuration can be written in one file (`configuration.nix`) and this file configures all of the programs that normally live in /etc and /bin and /usr/bin and such, although instead it symlinks the versions of the packages to use from the "nix store". All dependent files/programs are hashed and if the build process changes at all, the hash changes and the target is rebuilt so that systems are precisely identical (as best as possible). Programs that use the search path to find a program (eg. env, which) are the best in this situation as many system paths are quite different from usual:

```
[jared@vmh02s:~]$ which bash
/run/current-system/sw/bin/bash

[jared@vmh02s:~]$ which sh
/run/current-system/sw/bin/sh

[jared@vmh02s:~]$ which systemctl
/run/current-system/sw/bin/systemctl

[jared@vmh02s:~]$ which docker
/run/current-system/sw/bin/docker
```

And so on.

### How can we reproduce this?

1. On a NixOS machine, install Docker and ensure the user is in the `docker` group.
2. Clone WebODM:

```
git clone https://github.com/OpenDroneMap/WebODM.git
cd WebODM
```

3. Try to run WebODM normally:

```
./webodm.sh start --hostname x.x.x.x --port 8000 --db-dir /srv/webodm/db --media-dir /srv/webodm/media --default-nodes 1
```

(Note - `/srv/webodm/db` and `/srv/webodm/media` exist on my system and are mounted in SSD/HDD scratch space; your configuration may differ. Also, redacted internal IP address)

**Observed behavior:**

The problem occurs in the project scripts, for example:

```
/home/jared/WebODM/webodm.sh: /bin/bash: bad interpreter: No such file or directory
```

To proceed, I had to patch all WebODM scripts using `#!/bin/bash` to use:

```
[...]

PR #1854, opened 2026-03-16
Fix bash paths to allow usage on NixOS

jrddunbr:
This addresses #1853 if that change is desired.

Note: The contribution guidelines note that I should provide this warning - AI was used in the development of the issue and subsequently this PR, however I was able to confirm that the fix/workaround command was able to create a working WebODM installation in Docker, and I believe that the changes shown below are correct. Please feel free to confirm in your own test environments, and I can provide a hand-tailored NixOS `configuration.nix` for testing if you would like to test it in a virtual machine for the verification of this PR. I've run it on my local NixOS box to my satisfaction, and already have nodes processing data and providing results.

I don't particularly mind if this PR is rejected as I've already mitigated the issue locally; I'm submitting this purely for the benefit of the next NixOS user to come along this project.

### DIFF
--- a/dev-worker.sh
+++ b/dev-worker.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Launch a development worker so that changes in a dev
 # environment can be refreshed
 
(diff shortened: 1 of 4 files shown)


pierotofy:
Looks good to me, thanks for the contribution!

PR #1876, opened 2026-04-12
Fix import of .tiff files

pierotofy:
Comparison of .tif => .tiff leads to an import error for external assets.


pierotofy:
Also need to fix an issue with exporting certain multispectral RGX images

Issue #1913, opened 2026-05-03
Feature request: make native version easier to setup on remote headless machines


aleottantotto:
I would like to have an easier setup in the native version to  be run on a headless machine and be accessible (like the Docker one) at localhost:port

I found some old forum replies where there is a procedure which apparently was successful, but I think I messed up since the performance on the native version was much worse than the Docker on same machine with same dataset and same preset

This feature for me is a must to try to switch to the native version since I'm running it on a headless remote machine

Thank you :)

pierotofy:
You can now launch headless mode via `WebODM.exe --headless-mode` on Windows 👍

PR #1916, opened 2026-05-08
Always show share button even in desktop mode

pierotofy:
It can be useful to share on a desktop instance, for example for giving QGIS access to the map tiles. There's already a warning in place saying that private links are not visible to the outside.


pierotofy:
Also helps get to #1913

PR #1951, opened 2026-05-27
JOSM plugin title in url and in JOSM desktop (issue #1945)

### LINKED ISSUE #1945: JOSM /imagery command should use the task name as the imagery title
When using the OSM Digitise > JOSM button the `title` parameter passed to `http://127.0.0.1:8111/imagery` is `undefined`. It's expected this should be the task name.

DMZimo:
Fixes #1945 

Sets the `title` and `changeset_source` correctly with the task's title or with the task's ID if it lacks a title (i.e - an orthophoto that was made directly through ODM/ODX) using the existing `encodeURIComponent`.

Tested with JOSM 19555 on Windows 11 with the [brighton_beach](https://github.com/pierotofy/drone_dataset_brighton_beach) dataset.

### DIFF
--- a/coreplugins/osm-quickedit/public/main.js
+++ b/coreplugins/osm-quickedit/public/main.js
@@ -99,8 +99,9 @@ PluginsAPI.Map.addActionButton(function (options) {
         }
 
         function prepareJOSM() {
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @DMZimo, looks great to me! :+1:

PR #1955, opened 2026-05-28
Fix: Dashboard pagination respects PAGE_SIZE and allows UI-configurable page size (#1944)

### LINKED ISSUE #1944: Dashboard pagination UI should respect custom page size setting
When setting a custom PAGE_SIZE via `--settings` 's `REST_FRAMEWORK = { 'PAGE_SIZE': 100 }`, list of projects in the dashboard is updated, but the page buttons still assume the default page size, and therefore clicking the later page numbers will not work and show an error

Image

DMZimo:
Fixes #1944 

Implemented custom dashboard pagination.

## Backend

**Created:** `app/api/pagination.py`:

- Added `PageSizePagination` subclass that accepts `?page_size=N` per request with a hard server cap of 100, and includes `page_size` in the paginated response body.

**Edited:** `webodm/settings.py`:

- Registered `PageSizePagination` as `DEFAULT_PAGINATION_CLASS` in DRF settings.

## Frontend

**Edited:** `app/static/app/js/components/ProjectList.jsx`:

- Reads `page_size` from the API response instead of the hardcoded value.

**Edited:** `app/static/app/js/Dashboard.jsx`:

- Initializes `page_size` from localStorage on first load (mirroring the existing pattern used for `ordering`).

**Edited:** `app/static/app/js/components/Paginator.jsx`:

- Page size selector in toolbar with custom input capped at 100.

**Edited:** `app/static/app/js/css/Paginator.scss`:

- Added styles for the custom page size input row.

## Tests and notes:

Both `page_size` and `page` are reflected in the URL:

`http://127.0.0.1:8000/dashboard/?page=4&page_size=50&ordering=-created_at&search=test`

It doesn't break anything on my end and ommiting the values just falls back to whatever is already defined in the localStorage or the defaults, but it's a possible URL mutation so I thought I should note it just in case.

Tested with 2000+ empty projects (crerated a test is in ~~`/app/tests/test_seed_projects.py`~~(See comment below) and simply ran it twice).

image

image


DMZimo:
Removed the `test_seed_projects` script. Django fired it as parts of the checks but it was only intened for manual use.

Here it is if anyone want to run it or properly implement it.

```python
# Creates 1000 test projects for the 'admin' user.
# Run with:
# docker cp app/tests/seed_projects.py webapp:/webodm/ && docker exec webapp bash -c "cd /webodm && /webodm/venv/bin/python seed_projects.py"

import os
import django
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "webodm.settings")
django.setup()

from app.models import Project
from django.contrib.auth.models import User

user = User.objects.get(username="admin")
projects = [Project(owner=user, name="Test Project " + str(i)) for i in range(1, 1001)]
Project.objects.bulk_create(projects)
print("Done. 1000 test projects created.")
```

pierotofy:
Thanks @DMZimo 🙏 

Can you please:

 * Disclose which AI you've used and to what extent
 * Make sure the pagination logic does not change the entire REST API, but just the projects for which this pagination fix is written?
 * Remove the "Custom pagination" input option? There's a cap of 100 (good), so it will confuse users to be able to type a custom option.

pierotofy:
Closing for now, we can re-open once the points above get addressed.

PR #1956, opened 2026-05-28
Fix Share to DroneDB silently doing nothing

HeDo88TH:
Three bugs prevented the Share button from uploading anything:

1. share_to_ddb was decorated with @task (raw Celery), but the worker process only loads 'worker.tasks' and 'app.plugins.worker'. The task was never registered in the worker, so messages were silently discarded. Switch to run_function_async() like every other plugin (objdetect, contours, measure, cesiumion).

2. When the user chose 'Create new dataset in org X', ShareDialog sent tag='orgslug' (no slash). Registry's ToTag() requires exactly 'org/dataset' and threw FormatException -> HTTP 400. Fix: send orgSlug as a separate form field instead of abusing the tag field.

3. share_init() and share_commit() were called outside any try/except. Any exception left status=1 (Running) forever in the datastore, so the frontend polled endlessly showing 'Sharing'. Wrap the entire upload flow in a try/except that sets status=2 (Error) on failure.

### DIFF
--- a/coreplugins/dronedb/api_views.py
+++ b/coreplugins/dronedb/api_views.py
@@ -394,11 +394,11 @@ class ShareTaskView(TaskView):
 def share_to_ddb(pk, settings, files, tag=None, org_slug=None, dataset_name=None):
     import os, time
     from app.plugins import logger
-    from app.plugins import get_current_plugin
+    from app.plugins.functions import get_plugin_by_name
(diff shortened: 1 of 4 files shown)


pierotofy:
Is this ready for review?

HeDo88TH:
> Is this ready for review?

Not yet, i'm still testing unfortunately

PR #1959, opened 2026-06-05
Store email, first name, and last name for new OIDC users

jet86:
Store email, first name, and last name for new users authenticating via OIDC.

First and last name are pulled from the profile OIDC scope, and stored only if present.

### DIFF
--- a/app/views/oidc.py
+++ b/app/views/oidc.py
@@ -65,7 +65,7 @@ def oidc_login(request, provider_index):
         'response_type': 'code',
         'client_id': provider['client_id'],
         'redirect_uri': callback_url,
-        'scope': 'openid email',
+        'scope': 'openid email profile',
         'state': state,
     }
 
@@ -149,6 +149,9 @@ def oidc_callback(request):
         messages.warning(request, _('SSO login failed.'))
         return redirect(settings.LOGIN_URL)
 
+    given_name = claims.get('given_name')
+    family_name = claims.get('family_name')
+    
     # Check if it's authorized
     authorized = True
     if settings.OIDC_AUTH_EMAILS:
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @jet86 !

Can you explain the use case for storing personal information? Not against it, but would like to hear why something is useful before we include it.

A few things:
 * Requesting additional scopes and storing personal information should be optional (opt-out by default), via an additional `OIDC_` setting in `settings.py`. When opted-out, the `profile` scope should not be requested.
 * OIDC's `profile` scope can provide: name, family_name, given_name, middle_name. What are the differences and is the current choice of mapping correct?
 * Since username = email, do we need to store email twice?

jet86:
The storing of their name would allow it to be used throughout the interface for personalisation if given (and we already have the fields for storing it manually anyway). You're absolutely right that it should be optional though, so I'll work out how to do that through an additional setting as you've suggested (and only request the profile scope if needed).

I'm mapping `given_name` to `first_name` and `family_name` to `last_name` as those are the equivalent/synonymous fields. The OIDC profile attribute `name` usually stores either the entire name (generally first name + ' ' + last name) or a nickname. Middle name just stores someone middle name(s) if they have any, though at least in my experience I haven't generally seen that used in practice.

I believe given that there is an email field, we should be storing their email in it when available. This would then allow for things like email notifications to individual users to be implemented down the track etc.

My aim for a subsequent PR would be to (optionally) allow a mapping of OIDC attributes to groups, to allow auto-assignment of groups via OIDC. So starting by looking at this more basic change has at least helped me to start to wrap my head around how that would be implemented (and your questions here are helpful for that too); so even if you don't in the end want anything from this PR merged, it's at least helped on my side...

pierotofy:
> would allow it to be used throughout the interface for personalisation

Since it sounds like a future improvement, I would wait on merging this until it's more clear what these personalizations / e-mail notifications would look like. I don't want to add new code that could be useful in the future. Once we get there, this code is already written and can be easily included then. :+1: 

The automatic group assignment sounds useful, if you want to discuss details before implementation let's chat about it on discord.

PR #1970, opened 2026-06-30
OIDC Group Management

jet86:
Adds several options to manage WebODM users and groups via OIDC claims:

- The ability to optionally update a user's profile (name and email) from OIDC
- The ability to automatically assign a user to, or remove a user from groups based on values returned in specified OIDC claims
- The ability to optionally create groups that are missing in WebODM when returned in specified OIDC claims
- The ability to treat specific groups as "manually managed" in WebODM by marking them as ignored by OIDC so that users are not removed from them if not returned in OIDC claims


pierotofy:
Looks good to me, running copilot for a second opinion.

pierotofy:
Thanks @jet86 ! 👍

PR #1981, opened 2026-07-15
Update to Ubuntu 24.04

pierotofy:
Updates WebODM docker images to Ubuntu 24.04


pierotofy:
Meh, maybe some other day

PR #1989, opened 2026-07-21
Add Podman compatibility to webodm.sh and Compose files

GameAnimeCode:
## Problem
`webodm.sh` and several Compose files assume a Docker environment. Running the stack under Podman fails or misbehaves in a few places:

- `check_docker_compose` and `environment_check` require a Docker CLI. They don't fall back to podman-compose on a Podman-only host.
- GPU passthrough used Docker's `deploy.resources.reservations.devices` block. Podman doesn't support that block. It resolves GPUs through CDI syntax instead (`nvidia.com/gpu=all`) and only when podman-compose talks to Podman directly. A Docker CLI pointed at a Podman socket doesn't resolve CDI devices, and the failure is silent: the GPU node container starts with no GPU wired in.
- Bind mounts in `docker-compose.settings.yml` and `docker-compose.ssl-manual.yml` need `:z/:Z` SELinux relabel suffixes to be readable under Podman with SELinux enforcing.
- `run_tests` only checked for `/.dockerenv` to detect that it's running inside a container. Podman uses `/run/.containerenv` instead.
- The Compose version check crashed with SIGPIPE under `set -o pipefail` because `head -n1` closes the pipe early against podman-compose's multi-line version output.

## Solution
- Detect Podman and fall back to podman-compose when no Docker or compose-plugin shim is present.
- Pick the GPU compose file based on the resolved compose backend (`uses_podman_compose`), not just whether Podman is present. Added `docker-compose.nodeodm.gpu.nvidia.podman.yml`, which uses CDI device syntax.
- Added `:z/:Z` suffixes to the two affected bind mounts.
- Check `/run/.containerenv` alongside `/.dockerenv` in `run_tests`.
- Guard the version-check pipeline with `|| true` so the SIGPIPE no longer kills the script.

Docker-only setups keep their existing behavior. The new logic only branches when Podman is the resolved backend.

## Testing
1. Removed Podman-Docker compatibility layer.
2. Removed all containers, volumes, and images.
3. Ran `./webodm start --gpu` with preset `.env` data locations.
4. Processed a [dataset](https://github.com/OpenDroneMap/odm_data_helenenschacht/tree/main) with the `3D Model` preset.
5. Ensured GPU was used.

It should be tested in a Docker engine environment to confirm no regressions.

### DIFF
--- a/docker-compose.nodeodm.gpu.nvidia.podman.yml
+++ b/docker-compose.nodeodm.gpu.nvidia.podman.yml
@@ -1,15 +1,16 @@
 # Podman-compatible alternative to docker-compose.nodeodm.gpu.nvidia.yml.
 #
 # The Docker variant requests the GPU via the Compose Spec's
-# `deploy.resources.reservations.devices` block (driver: nvidia), which relies
-# on the nvidia-container-toolkit runtime and isn't understood by Podman's
-# Docker-API compatibility layer. Podman instead exposes GPUs through the
-# Container Device Interface (CDI), referenced here the same way this
-# project's own devcontainer does it (see .devcontainer/devcontainer.json's
+# `deploy.resources.reservations.devices` block (driver: nvidia), which needs
+# the nvidia-container-toolkit runtime and isn't understood by Podman. Podman
+# resolves GPUs through the Container Device Interface (CDI) instead, the same
+# way this project's own devcontainer does it (see .devcontainer/devcontainer.json's
 # `--device nvidia.com/gpu=all`).
 #
-# webodm.sh selects this file instead of docker-compose.nodeodm.gpu.nvidia.yml
-# when it detects the active container engine is Podman.
+# webodm.sh picks this file over docker-compose.nodeodm.gpu.nvidia.yml when
+# podman-compose is the resolved compose backend (see uses_podman_compose()
+# in webodm.sh; CDI resolution needs podman-compose talking to Podman
+# directly, not just any engine running under Podman).
 version: '2.1'
 services:
   webapp:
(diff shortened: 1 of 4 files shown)


GameAnimeCode:
Related Issue: https://github.com/WebODM/WebODM/issues/1987

GameAnimeCode:
Associated Docs PR: https://github.com/WebODM/Docs/pull/9

pierotofy:
This is the excessive complexity I was talking about. 

No additional yaml files please. No ad-hoc logic for podman. If we can make it work for both docker and podman, great. If not, we're not going to support two container engines. :pray: 

Also:

> It should be tested in a Docker engine environment to confirm no regressions.

This should be something you test. Did you use AI to make these changes? Do you understand each change?

GameAnimeCode:
Thanks for the feedback. To clarify the extra compose file: it's not ad-hoc logic; it's a difference in how the two engines expose a GPU to a container. Docker reserves it through `deploy.resources.reservations.devices` with the nvidia-container-toolkit runtime. Podman doesn't support that block at all. It resolves GPUs through CDI instead, with a device entry like `nvidia.com/gpu=all`. There's no single compose syntax that expresses both, so a shared file isn't an option here, only one engine or the other.

Without the second file, GPU passthrough under Podman doesn't fail loudly; it just starts the node with no GPU attached, which seemed worse than adding the extra file. `webodm.sh` only picks it when podman-compose is actually resolved as the backend, so Docker-only setups never see it.

Happy to verify against a plain Docker environment to confirm no regressions there. If the project would rather not carry a second engine at all, I understand. I can maintain this as a Podman-focused fork instead.

Yes, I used AI assistance for parts of this, which is fine under the CONTRIBUTING & AI policy, and I reviewed and tested every change myself.

pierotofy:
Appreciate the effort! Yes, I don't want to support multiple container engines this way. It's a tremendous headache just to support one.

> I can maintain this as a Podman-focused fork instead.

:+1:

PR #1993, opened 2026-07-27
measure: split volume into cut/fill/net, add DSM/DTM model selection

thadwald:
The measure plugin previously calculated volume as the absolute value of the sum of cut and fill. It also incorrectly used the Digital Surface Model data for this calculation, even when the user had selected DTM as the visible layer. 

Now the Measure results show

- which model is being measured (DSM or DTM)
- Cut (material below the plane)
- Fill (material above the plane)
- Net (fill - cut (signed) ) 

summary of changes to files: 

MeasurePopup now determines which model to use and sends it with the volume call. it also displays all of the new information listed above. 

dem_type parameter ("dim"/"dtm")  was added to TaskVolume, which verifies that the resource exists and sends it with the calc_volume call.

calc_volume now calculates and returns cut, fill and net individually

### DIFF
--- a/coreplugins/measure/api.py
+++ b/coreplugins/measure/api.py
@@ -13,23 +13,31 @@ from .volume import calc_volume
 class VolumeRequestSerializer(serializers.Serializer):
     area = serializers.JSONField(help_text="GeoJSON Polygon contour defining the volume area to compute")
     method = serializers.CharField(help_text="One of: [plane,triangulate,average,custom,highest,lowest]", default="triangulate", allow_blank=True)
+    dem_type = serializers.ChoiceField(choices=["dsm", "dtm"], help_text="Which elevation model to compute the volume against", default="dsm")
 
 class TaskVolume(TaskView):
     def post(self, request, pk=None):
         task = self.get_and_check_task(request, pk)
-        if task.dsm_extent is None:
-            return Response({'error': _('No surface model available. From the Dashboard, select this task, press Edit, from the options make sure to check "dsm", then press Restart --> From DEM.')})
 
         serializer = VolumeRequestSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
         area = serializer['area'].value
         method = serializer['method'].value
-        points = [coord for coord in area['geometry']['coordinates'][0]]
-        dsm = os.path.abspath(task.get_asset_download_path("dsm.tif"))
+        dem_type = serializer['dem_type'].value
 
-        try: 
-            celery_task_id = run_function_async(calc_volume, input_dem=dsm, pts=points, pts_epsg=4326, base_method=method).task_id
+        if dem_type == "dtm":
+            if task.dtm_extent is None:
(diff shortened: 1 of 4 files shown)


pierotofy:
Can you show one or two screenshots showing your UI changes and explain how did you test to verify this implementation is correct?

thadwald:
The test project involves a small pea rock fall protection pit surrounding a play structure at a school. The pit is bounded on three sides with a cement curb with the 4th side unfinished for dump truck access. I flew and modeled the site to determine the amount of pea gravel needed. 

The original WebODM report indicated I need about 75 cubic yards, which comes neatly to 100 tons at 1.3 tons per cu.yd, or four truck loads. I noticed after a few loads that this would be short and added one more load, which filled the area. 

So ODM estimation was short by about 20%.

The original number was 75 cu.yds.. I did not like that it was positive. I also saw that WebODM Lightning has a cut/fill illustration, so I tasked AI to add cut, fill, and signed Net volume. 

The resulting numbers indicated why the estimate was off by 20%; it was using the surface model, which included the volume of the structure in the calculation and subtracted it from the estimate. 
Screenshot 2026-07-27 at 7 30 08 AM

So I tasked AI to pass along the active layer as the target DEM. The resulting Fill is now correctly near zero and the resulting volume measurement closely matches the actual required volume for this job. 
Screenshot 2026-07-27 at 7 31 59 AM

I tested it on a model without a DEM; it displays the "model not found" error in the flyout box, albeit with the long text flowing out of the box ... but this would be a change for a different PR. I also tested it on some stockpiles and quickly saw why this feature has little use in stockpile measurement. ( 1.) default DTM settings results in the the tops of stockpiles being chopped off, 2) Cut is always near zero, and 3) Stockpiles are always a positive volume). But it does no harm in this use case AND it is quite useful for non-stockpile applications, as I demonstrated here.

One design question that I had was; should the operation calculate the volume for all models available? I'm leaning towards "No" but it would be great to have more experienced minds weigh in.

pierotofy:
Thanks for the screenshots and explanation of your dataset. But how did you *verify* that the output numbers are correct? (Do you understand the changes AI has written? How do you know they are correct?)

As an implementation note, you don't want the program to pick a default surface, you want to know if there are multiple surfaces available (none, DSM, DTM, DSM+DTM ?) under the area of the polygon (but remember, there could be multiple tasks in Project View!), if you have more than one surface then display a selector (defaulting to DSM).

Please do not simply copy paste this into an AI though; it does not help to just prompt additions to WebODM (I can do that too and it takes me 2 seconds). The real work is in validating these changes and testing (and if you can do that, then it helps, but if you can't, let's leave this as a "proof of concept"). 🙏

thadwald:
So I do have at least some programming experience and, yes, I understand the changes that AI has written. 
What I lack is an intuition for how WebOMD works. For instance, I didn't think of the fact that a project can have multiple tasks, so I didn't test or even consider what happens in that case. This issue of not being familiar with the project is probably going to persist. 

You have given me some direction and I will keep working on it as I find time and in the meantime I have changed this pr to a draft.

PR #2010, opened 2026-08-25
Orthophoto lighting fix

n-grieger:
This fix addresses the coloring issues of our orthophotos mentioned in https://github.com/OpenDroneMap/ODM/issues/2064. Apparently, the orthophoto viewer calculates image statistics (including min and max values of the color histogram) across all available bands in the orthophoto despite only displaying the RGB bands. This can lead to very high coloring max values and correspondingly very dark orthophotos on default settings (see histogram in the screenshot below, the orange distribution is not an RGB band).

Old version, orthophoto viewer:
image

To fix this, I reduced the channels for the statistics calculation to the RGB bands in `app/api/tiler.py`. In addition, I switched to the percentile based min/max color value calculation already used for the plant health viewer (applied to both the orthophoto viewer and the thumbnail on the Dashboard). Finally, I fixed a small bug in the calculation of the min and max values of the histogram, which were based only on the last band in the loop rather than all input bands (see `app/static/app/js/components/Map.jsx` and `app/static/app/js/components/LayersControlLayer.jsx`). Our test deployment now shows a much lighter orthophoto and only the three RGB bands in the color histogram.

Fixed version, orthophoto viewer:
image

Open issue: With this fix, we lose information about non-RGB bands in the color histogram of the orthophoto. An alternative to the changes proposed here would be to show all bands in the histogram but using only RGB bands for min/max calculation.

### DIFF
--- a/app/api/tasks.py
+++ b/app/api/tasks.py
@@ -628,7 +628,7 @@ class TaskThumbnail(TaskNestedView):
         if img.dtype != np.uint8:
             img = img.astype(np.float32)
 
-            # Match the tiler's percentile rescale (tiler.py pmin/pmax=2/98) so the thumbnail and map view stretch identically
+            # rescale RGB between 2nd and 98th percentile, ignore alpha
             if img.shape[2] == 4:
                 valid = img[:,:,3] > 0
             else:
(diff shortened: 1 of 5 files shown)


pierotofy:
Closing in favor of https://github.com/WebODM/WebODM/pull/2029

Thanks for the draft!

Issue #2027, opened 2026-09-21
Console output sync stalls permanently if the node's line count goes backwards


LuisPCFialho:
# Console output sync stalls permanently if the node's line count goes backwards

## Summary

`Task.process()` asks the processing node for console output starting at an
offset it derives from the console it has already stored. If the node's output
ever becomes *shorter* than that offset, NodeODM returns an empty list with no
error, the stored console never grows, the offset never changes, and console
sync is stuck permanently — while the task itself keeps processing normally.

The UI gives no indication of this. To the user the task looks frozen: the
console stops at a timestamp minutes or hours in the past, while `processing_time`
keeps ticking up.

## Where

`app/models/task.py` (still present on `master` as of 2026-09-21):

```python
if not self.console.output():
    current_lines_count = 0
else:
    current_lines_count = len(self.console.output().split("\n"))

info = self.processing_node.get_task_info(self.uuid, current_lines_count)
```

The offset is derived purely from local state and is monotonically
non-decreasing. Nothing validates it against the node.

`self.console.reset()` exists, but is only called on the `RESTART` pending
action path, i.e. only when WebODM itself initiates the restart.

## Why the node can go backwards

NodeODM's `/task/<uuid>/output?line=N` returns the lines from `N` onwards of
the **current** run. Verified on NodeODM (WebODM 2.x desktop, node on
`127.0.0.1:29301`), on a task with ~7000 lines of output:

```
line=0       -> 6988 lines
line=5000    -> 1988 lines
line=999999  ->    0 lines      <- no error, empty list
```

So any event that makes the node's current output shorter than WebODM's stored
console produces the stall. Ways this happens in practice:

- the task is re-run on the node (`--rerun-from`), which resets its output
- WebODM's own processes restart mid-task and a restart is issued around the
  same time, so the `RESTART` pending action's `console.reset()` does not run
  but the node does restart the task
- the node's task data is recreated

## Observed

WebODM 2.x desktop on Windows, ODM 3.x, single local NodeODM, one 412-image
task.

```
node (doing the work)      last line 16:14:52   <- live
WebODM (showing it)        last line 16:06:59   <- frozen, 8 minutes behind
```

`processing_time` kept increasing the whole time (51.3 min and rising), and
`running_progress` kept being written, so only the console was affected — which
makes it harder to notice, because the task does not look dead.

Manually truncating the stored console made sync recover on the next cycle
(~5 s), and the console refilled from the current run's line 0.

## Suggested fix

[...]

PR #2028, opened 2026-09-21
Add Real Progress coreplugin: true percentage and ETA on the task list

LuisPCFialho:
Adds a coreplugin that puts the true percentage and the remaining time of each
running task on the task list.

## Why

`running_progress` only moves at stage boundaries. A task sits at 4% for two
hours and then jumps. There is no way to tell whether matching is half done or
nearly finished, and no way to know when the task will end.

The fine-grained progress already exists — it is in the console, one line per
matched pair, per processed image, per image added to the reconstruction.

Before:

```
Running
```

After:

```
6.3% · Camera positions · 11,298 of 84,666 · remaining 2h13
```

The tooltip lists all 13 ODM stages with their state, the counts and rates of
each sub-step, and the estimated finish time.

## How

Three sources, in order of reliability:

1. Stage markers (`Running X stage` / `Finished X stage`). These always say
   exactly where the task is and which stages are done.
2. Step lines inside OpenSfM — features, matched pairs, images added to the
   reconstruction, undistorted images. This is where the fine percentage and the
   rate come from.
3. Files on disk for stages that log no progress: OpenMVS writes one `.dmap` per
   image and says nothing, so those files are counted. Local node only; the
   task's working directory is taken from the console itself.

Stage weights for the overall percentage start at typical values and are
replaced by measured durations as stages complete, so the estimate corrects
itself as the task goes on.

Reading is incremental — a line cursor per task, only new lines requested — with
a 4 s floor between node queries and a 5 s client poll. A large task's console
goes past 20 MB, so it is never re-read whole except in the one recovery case
described below.

## It also works around #2027

`Task.process()` derives the console offset from the console it already stored.
If the node's output becomes shorter than that offset — which happens when a
task is re-run on the node — NodeODM returns an empty list with no error and
console sync stalls permanently while the task keeps processing. Filed as #2027.

This plugin reads the console from the node, so it is unaffected. It also
detects the stall (the node is producing new lines while the stored console has
not been touched for minutes) and calls `console.reset()` so normal sync
resumes. If #2027 is fixed, that code becomes dead weight and can be dropped —
happy to remove it if you would rather fix it upstream first.

## Notes

- No configuration, no settings page, no menu entry. It applies to every running
  task.
- Strings are translated server side following the active language. 19 languages
  are included; anything else falls back to English. I did not use gettext here
  because it would mean compiling `.mo` files inside the plugin folder for
  thirteen short strings — happy to switch to gettext if you prefer consistency
  with the rest of the codebase.
- Rows are matched to tasks by name, which is the only identifier present in the
  DOM while a task is running.
- Stage markers were verified against ODM 3.5.5 logs, including a completed run.
  Unknown stages degrade to the stage name alone, never to an error.
- `main.js` is plain ES5, no JSX build step. It repaints on each poll and on DOM
  mutations because React re-renders the rows, and picks the most detailed text
  that fits the available width by measuring it.

Also published standalone at
https://github.com/LuisPCFialho/webodm-real-progress for anyone who wants it as
a zip.


pierotofy:
Thanks for the idea! I've opened https://github.com/WebODM/ODX/issues/49 to track it. While the intent is clear, I don't think this is the right way to go about implementing it (and AI doesn't either).

